PR #73, opened 2019-12-08
Added ntpreview to command list

HexedHero:


NEZNAMY:
Sure why not, the help menu needs update anyway.

PR #74, opened 2019-12-09
Change section sign encoding to UTF-8

pgmann:
Just some very minor fixes I thought you might appreciate that make developing/building on a different machine easier:

* Changed `§` from platform-specific character to UTF-8 encoding version
* Fixed typo: `occured` -> `occurred`
* Added `.idea` to the gitignore to prevent IntelliJ project files getting committed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ hs_err_pid*
 .classpath
 .project
 /bin/
+
+# intellij
+/.idea/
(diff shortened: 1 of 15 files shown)


NEZNAMY:
Github also says you made a change to the § character, however i wasn't able to find out what the difference is. I tried to run 
`System.out.println((int)'§');
System.out.println((int)'§');`
with both previous and new character, both of these returned 167.

pgmann:
Basically when I cloned the project all the characters displayed as the `?` missing character and IntelliJ asked if I wanted to switch encoding to a windows-specific one instead of UTF-8. Instead I replaced all instances of the `?` character with `§`.

The difference is `�` vs `§`. The first one is a `\ufffd` (replacement character) in UTF-8, while the second is `\u00a7` which is correct.
image

NEZNAMY:
I'm going to have fun manully merging this with new changes i did not commit yet.

PR #132, opened 2020-04-08
I want to be able to use getDataFolder

130B:
This allows you to use custom plugin paths, including configuration files

### DIFF
--- a/src/me/neznamy/tab/platforms/bukkit/features/ExpansionDownloader.java
+++ b/src/me/neznamy/tab/platforms/bukkit/features/ExpansionDownloader.java
@@ -28,7 +28,7 @@ public class ExpansionDownloader{
 							Main.usedExpansions.removeAll(PlaceholderAPI.getRegisteredIdentifiers());
 							Shared.debug("Expansions to install: " + Main.usedExpansions);
 							if (!Main.usedExpansions.isEmpty()) {
-								File expansionsFolder = new File("plugins" + File.separatorChar + "PlaceholderAPI" + File.separatorChar + "expansions");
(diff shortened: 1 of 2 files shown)


NEZNAMY:
The ConfigurationFile commit (https://github.com/NEZNAMY/TAB/pull/132/commits/3e94e13228480a2d3d901e290dd827c9a3031b79) would result in NoClassDefFoundError on bungeecord and velocity. The class is universal and i want to keep it like that.

130B:
> The ConfigurationFile commit ([3e94e13](https://github.com/NEZNAMY/TAB/commit/3e94e13228480a2d3d901e290dd827c9a3031b79)) would result in NoClassDefFoundError on bungeecord and velocity. The class is universal and i want to keep it like that.

There is also getDataFolder in BungeeCord, Maybe it can read the plugin path and pass parameters to that class as soon as the server is loaded.

NEZNAMY:
That's possible, but.. why ?

130B:
For many duplicate sub-servers, it is not necessary to copy every directory, which can reduce unnecessary trouble

NEZNAMY:
How do you even make .getDataFolder return something else than it does now ?

A248:
There is a spigot startup flag which no one ever uses to change the /plugins/ directory.

NEZNAMY:
I will consider it when i get more requests.

NEZNAMY:
Both of the changed classes contain critical issues.
PlaceholderAPI downloader class points to wrong directory.
ConfigurationFile class would stop working on bungeecord.
I will add this myself.

PR #235, opened 2020-11-04
Fixed ViaVersion dependency

asineth0:
Changed 3.2.0 to 3.2.1-SNAPSHOT, since 3.2.0 will result in a 404 while building. It builds fine now.


NEZNAMY:
You broke velocity path, making it no longer universal.

asineth0:
Hold on, I'll make another PR, I figured out the issue.

PR #240, opened 2020-11-14
Fixed ArrayIndexOutOfBoundsException in conditions

Tanguygab:
When using Equals and NotEquals conditions types and the rightside is empty


Tanguygab:
why do I suck so much with GitHub

PR #281, opened 2021-02-27
Fix using wrong cpu usage deserialization packet type

msengbusch:
I think there's a typo in the onHeaderFooter method in FeatureManager.java. This leads to that the cpu usage for deserialization will be calculated for the wrong type of packet.

https://github.com/NEZNAMY/TAB/blob/9f3995198b2d6bd84d188f5a93bb78fb785e6ee7/src/main/java/me/neznamy/tab/shared/FeatureManager.java#L330-L333

I changed it to the packet header and footer usage type, so that the cpu usage will be correct.


NEZNAMY:
Nice catch, thanks.

PR #322, opened 2021-06-08
Added support for judgment of world or server.

Score2:
E.g:
```
disable-in-worlds:
-'example-*'
```
Then the player will not enable this feature in example-1 or example-2.

### DIFF
--- a/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/BukkitPlatform.java
+++ b/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/BukkitPlatform.java
@@ -1,11 +1,10 @@
 package me.neznamy.tab.platforms.bukkit;
 
 import java.io.File;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.*;
 
 import org.bukkit.Bukkit;
+import org.bukkit.World;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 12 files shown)


NEZNAMY:
I was wondering why not just adding some logic into https://github.com/NEZNAMY/TAB/blob/master/shared/src/main/java/me/neznamy/tab/shared/features/types/Feature.java#L22-L35, sounds like that would do it just fine to me.

Score2:
> ‎我想知道为什么不只是添加一些逻辑‎[‎到 https://github.com/NEZNAMY/TAB/blob/master/shared/src/main/java/me/neznamy/tab/shared/features/types/Feature.java#L22-L35，‎](https://github.com/NEZNAMY/TAB/blob/master/shared/src/main/java/me/neznamy/tab/shared/features/types/Feature.java#L22-L35)‎听起来像那样会做它只是罚款给我。应要求加载整个世界/服务器列表看起来没有必要。‎

Adding a group of servers is to accommodate when you are facing a large number of halls.
The numbers of these halls are usually uncontrollable. Sometimes there are many hall servers. It is not elegant to fill in the configuration file one by one, and it is very difficult to manage.

Score2:
Maybe that part of the logic I wrote is not perfect. If so, can you add the requested feature?

NEZNAMY:
Implemented in a much cleaner way.

PR #323, opened 2021-06-09
Support to obtain bukkit permissions from bungee.

Score2:
Use the luckperms plugin to use TAB permission control in bungee mode, the permission must be controlled by bukkit
Otherwise, once the tab.group.xxx permissions are set, the entire server will spread, even if the permissions are set to the "server=" function of the luckperms plug-in, it is useless


NEZNAMY:
This is already present, just for groups through vault, not permission nodes which no one should use anyway.

NEZNAMY:
Please resolve the conflict and I will merge this.

PR #359, opened 2021-07-27
Add %mspt%

ItzRiotz:
Only for PaperSpigot or forks of it.
I just did this for my server because I wanted to have it, but I can contribute if it's good enough? lol
Tried making it as compatible as possible for contribution and possible use of it.
Only displays current MSPT

### DIFF
--- a/bukkit/pom.xml
+++ b/bukkit/pom.xml
@@ -39,9 +39,20 @@
 			<id>aikar</id>
 			<url>https://repo.aikar.co/content/groups/aikar/</url>
 		</repository>
+
+		<repository>
+			<id>papermc-repo</id>
+			<url>https://papermc.io/repo/repository/maven-public/</url>
+		</repository>
 	</repositories>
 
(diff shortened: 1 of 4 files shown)


NEZNAMY:
How about replacing all of that with just
```
try {
	Class.forName("com.destroystokyo.paper.PaperConfig");
	placeholders.add(new ServerPlaceholder("%mspt%", 1000) {
		public String get() {
			try {
				return String.valueOf(roundDown.format(Bukkit.getAverageTickTime()));
			} catch(Exception e){
				return null;
			}
		}
	});
} catch (Exception e){
    //not paper
}
```
inside registerPlaceholders method?

ItzRiotz:
Didn't think of that :)
Sure, it's much simpler to the way I approached it

NEZNAMY:
Do you want to change the PR or should I just add it?

ItzRiotz:
I'll change the PR

PR #366, opened 2021-08-04
Update README.md

etil2jz:
Since TAB.java has been modified, we need to fix the line once again


NEZNAMY:
Quite sad to see this as the only type of contribution.

Tanguygab:
~~but NEZNAMY, this is super important! how will we know how to get the premium version for free then?~~

PR #376, opened 2021-08-13
Set velocity version to 3.0.0

DolphinGamerYT:
I was looking to update the velocity version to 3.0.0 for a client, but switching the version from the maven file all seems to work fine.

### DIFF
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -40,7 +40,7 @@
 		<dependency>
 			<groupId>com.velocitypowered</groupId>
 			<artifactId>velocity-api</artifactId>
-			<version>1.1.0</version>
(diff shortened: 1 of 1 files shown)


NEZNAMY:
Does this accomplish anything?

DolphinGamerYT:
> Does this accomplish anything?

For some reason it wasn't working with 3.0.0, but just changing the version to that it started working. It has something to do with how the plugin initializes.

NEZNAMY:
Can you send the error? It starts just fine for me.

DolphinGamerYT:
I just realized that maybe the issue was caused because I was using TAB 2.9.0

NEZNAMY:
Yes, that's the cause.

PR #389, opened 2021-09-20
Update readme file

Kamilkime:
• Add XML syntax to the Maven code block, so it looks nicer
• Add `provided` scope to Maven dependency, to make sure people don't compile TAB into their plugins (by default Maven/Gradle treat undeclared scope as `compile`)


NEZNAMY:
That explains why we got so many people complaining the API doesn't work, which turned out to be caused by class shading.

PR #432, opened 2021-12-08
Load TAB after PremiumVanish

LeonMangler:
When BungeeTabPlayer#isVanished uses PV's API, PV needs to actually be enabled, not just loaded. A soft-dependency should ensure that PV is fully enabled before TAB's onEnable() uses the API. It doesn't look like Bungee's PluginManager has a Plugin#isEnabled method.
This should fix [this error](https://pastebin.com/5B0ia4Vc ) that someone reported to me.

### DIFF
--- a/bungeecord/src/main/resources/bungee.yml
+++ b/bungeecord/src/main/resources/bungee.yml
@@ -4,4 +4,4 @@ version: @version@
 author: NEZNAMY
 description: @description@
 depends: []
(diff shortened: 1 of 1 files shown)


NEZNAMY:
This error, however, comes from a refreshing task, not onEnable as you can see in the stack trace. I can approve this, but I don't think it will solve the issue.

PR #452, opened 2021-12-22
Setup workflow artifact

kacperleague9:
As in title, hope you like it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,4 +41,4 @@ jobs:
         uses: actions/upload-artifact@v2.3.1
         with:
           name: Jar
-          path: /bin/**.jar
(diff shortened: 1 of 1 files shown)


BomBardyGamer:
Thanks for the pull request! As nice as this may be to have, I am concerned that this will allow pre-release versions to be downloaded by the public, when they should only be downloadable by buyers, as that's one of the perks of being a buyer.

kacperleague9:
Everyone can do that anyway. But for some people with genuine plans (ex. testing compabilities) have problems compiling it.

kacperleague9:
> What problems do you have compiling the plugin? It might be better if you report these issues on our Discord and I can help you sort them out there.

with this plugin, i haven't ran into any, but there are plugins which my computer had problem compiling

Issue #430, opened 2021-12-08
Placeholders break gradients


Faun471:
**Server version**
Purpur 1428 (Version 1.17.1)

**Plugin version** ("latest" is not a version)
TAB 3.0.0

**Describe the bug**
Using placeholders inside of gradient tags `<#HEX></#HEX>` will break the placeholder. I believe the attached images should be enough.

**How to reproduce**
Do `/tab parse me <#FBD72B>Players Online: %staffonline%/75</#F9484A>`
![image](https://user-images.githubusercontent.com/41022228/145149145-d9cef1c9-248b-469d-a539-6a66b049cc7b.png)

**Expected behavior**
The placeholder will properly be parsed, and the gradient won't break.
![image](https://user-images.githubusercontent.com/41022228/145149316-98c09c8f-d60d-45b0-a0e5-60928ee206dd.png)

**Actual behavior**
![image](https://user-images.githubusercontent.com/41022228/145149348-59801c04-8bcc-4589-9bd3-79bf56fa1411.png)

**Additional info** (errors, screenshots, ...)
Plugins (4): Essentials, PlaceholderAPI, TAB, Vault

PR #552, opened 2022-04-05
Add support for Kyori MiniMessage RGB format

lucypoulton:
`<#rrggbb>`


NEZNAMY:
The reason why this format is not supported is that there's a gradient pattern starting the same way, which caused problems (https://github.com/NEZNAMY/TAB/issues/430) when pre-parsing hex colors to save resources (https://github.com/NEZNAMY/TAB/blob/master/shared/src/main/java/me/neznamy/tab/shared/DynamicText.java#L113) when refreshing to only perform this operation once, instead of every placeholder refresh (if applicable). So if a placeholder was inside gradient, it wasn't parsed yet, but the format was, which then caused problems.

In order to accept this, pre-parsing must either be disabled completely, or at least formats must be disabled and only gradients allowed.

NEZNAMY:
Seeing more issues coming from current way of pre-processing in general, disabling rgb formats by adding the line I suggested will solve some problems as well. Once you add it, I can merge this.

PR #568, opened 2022-04-15
Add a per-permission header/footer

ruViolence:
Adds the ability to set header/footer when a player has a certain permission.

### DIFF
--- a/bukkit/src/main/resources/bukkitconfig.yml
+++ b/bukkit/src/main/resources/bukkitconfig.yml
@@ -28,11 +28,12 @@ header-footer:
         - "This is a shared header for"
         - "world2 and world3"
   per-permission:
-    some-tab:
+    any-tab-name:
+      permission: "some.tab"
       header:
         - "To see this tab you need a"
       footer:
-        - "'some-tab' permission"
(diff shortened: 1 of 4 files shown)


Tanguygab:
I think it could be even more interesting if you use conditions instead of just permissions, it could allow more things to be checked like regions and stuff =)

ruViolence:
This is quite difficult to do, given that the tab is updated only when the player is connected to the server or change the world. And I don't want to fill this plugin with my spaghetti code. In any case, you can do it yourself if you need it :)

Tanguygab:
true, maybe I'll try to look into it

NEZNAMY:
I don't want any new features in the plugin. On top of that, permissions are usually separated with `.`, just like config sections, which will cause conflict.

ruViolence:
Fixed a problem with dotted permissions. I think this feature would be useful for some people. On my server I use it to display a special header with a technical information about server performance to the staff players.

NEZNAMY:
This change would make the configuration more complicated for everyone, but you are currently the only one interested in this. I believe this change made for your server should not cause confusion of everyone else.

PR #582, opened 2022-04-29
Removable lines from header and footer

ruViolence:
Adds the ability to remove lines from a tab using a special value "#rm#". This will allow making various dynamically appearing lines using custom placeholders.


NEZNAMY:
You already can use dynamic lines using "\n", which sounds easier to use as well.

ruViolence:
Hmm, actually yes, but then I have to add a placeholder at the end of some line above, which might look a bit ugly in the config 😅

NEZNAMY:
And now it will look ugly in both config and code.

ruViolence:
To be honest, I did not realized right away that I could use \n to make this feature possible. Okay then, since it also adds some performance loss, I'll close this pull request.

PR #620, opened 2022-06-15
Add option to hide scoreboard line when its empty

Booquefius:
This just adds a boolean to allow users to hide a scoreboard line if its fully empty. This is useful when doing HCF styled scoreboards.


BomBardyGamer (approved):
Looks good to me

NEZNAMY:
This pull request does nothing.
This function is already present and this PR doesn't change anything, considering team name is never empty.

Booquefius:
Yea, the feature seems to be undocumented then. Would make sense why it works in testing if the teamName is never empty.

PR #629, opened 2022-06-23
Remove unnecessary team registration

efekurbann:
TAB was overriding the teams regardless the setting "anti-override". It only should do that when unlimited mode is disabled.
<br>
Tested with unlimited mode both enabled and disabled, does not seem to cause any problem.


Tanguygab:
not using teams when the unlimited nametag mode is enabled wouldn't allow sorting

NEZNAMY:
Teams are also used to make original name invisible, which users typically don't have another plugin for.

PR #632, opened 2022-06-26
Returns support Velocity

HarvelsX:
I like your plugin, but I use Velocity in a proxy server role.
The plugin already has Velocity implementation, but it is disabled by default, so why?
I updated Velocity API and replaced deprecated methods.
It looks a little strange to use your own implementation of the profiling tool, for what?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -12,7 +12,7 @@ val platforms = setOf(
     projects.bukkit,
     projects.bungeecord,
 //    projects.krypton,
-//    projects.velocity
+    projects.velocity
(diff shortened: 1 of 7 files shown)


Tanguygab:
Velocity support has been dropped due to a lack of required API methods for most features (especially everything regarding scoreboard teams) and was just a pain to maintain.
If you want to use TAB while still using Velocity, we'd recommend you use TAB on your backend servers instead.

HarvelsX:
Okay, I think I hear you. Thank you

PR #662, opened 2022-07-26
Update Gradle Wrapper to 7.5

JT122406:


BomBardyGamer:
I'll get this merged as soon as I fix the current reason why it's not building, which is a me issue.

PR #697, opened 2022-08-27
Updates and Velocity Deprecation Fix

JT122406:
Yes I know its not in the jar but still nice to keep it up to date for people wanting to use it by compiling themselves


NEZNAMY:
Looks like you need to fix compilation.

JT122406:
Yeah for some reason one of the repos are down

JT122406:
Let me see whats going on because it compiled for me earlier

JT122406:
All good now

PR #739, opened 2022-11-07
bump org.yaml:snakeyaml to 1.33

CoachYT1:
snakeyaml 1.30 is affected by 5 vulnerabilities and should be upgraded to the latest version


NEZNAMY:
TAB does not include snakeyaml library, only depends on it. It is brought to the system by the server jar, so it should be updated there. Assuming the API didn't change, this will have no effect at all.

PR #854, opened 2023-03-27
reduced number of threads in the pool, fixed races

zelvalea:
I'll start with the fact that the total number of threads in the pool is justifiably large, a thread is a rather expensive resource that is better to save (especially since most wheelbarrows have less than 16 threads) you have 20

Race:
When we read / write to fields that are updated / read by different threads, we must coordinate the memory behavior - this requires volatile,

more details: https://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html

same goes for ArrayList, it's not thread-safe, which means the state between submit and enable() is transitively broken)

and for dessert, in a user case, when we often write (in different threads) and rarely read, it is better to use LongAdder, it provides less competition for threads

P.S I did not check what I wrote, but I think you will figure it out)

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/CpuManager.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/CpuManager.java
@@ -3,12 +3,14 @@ package me.neznamy.tab.shared;
 import java.util.*;
 import java.util.Map.Entry;
 import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.LongAdder;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 import me.neznamy.tab.api.feature.TabFeature;
 import me.neznamy.tab.api.task.ThreadManager;
+
+
 
 /**
  * A class which measures CPU usage of all tasks inserted into it and shows usage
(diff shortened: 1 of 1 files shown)


NEZNAMY:
4 threads is way too low. The plugin creates 4 threads just for repeating tasks. You also need some for delayed tasks.

zelvalea:
> 4 threads is way too low. The plugin creates 4 threads just for repeating tasks. You also need some for delayed tasks.

in fact, this is quite enough, no one blocks in the pool and does not wait for pending tasks, you can read the implementation of ThreadPoolExecutor (which inherits ScheduledThreadPoolExecutor) in detail.  If, under some circumstances, 4 threads are really not enough for you, you can use Executors.newCachedThreadPool

zelvalea:
> > 4 threads is way too low. The plugin creates 4 threads just for repeating tasks. You also need some for delayed tasks.
> 
> in fact, this is quite enough, no one blocks in the pool and does not wait for pending tasks, you can read the implementation of ThreadPoolExecutor (w
![(https://user-images.githubusercontent.com/49918694/228159721-fd3d33a6-f963-4a22-857f-0a16ce5f1d80.png)
hich inherits ScheduledThreadPoolExecutor) in detail. If, under some circumstances, 4 threads are really not enough for you, you can use Executors.newCachedThreadPool

here is a picture of the profiler, almost all threads are sleeping

PR #856, opened 2023-03-28
optimize

zelvalea:
made several "optimizations"

in the previous PR, I changed AtomicLong to LongAdder because we had a lot of writers, in this case, when we have no competition for writers (we write only in one thread), LongAdder can, on the contrary, reduce performance and increase memory consumption

in CHM merge/compute guarantees an atomic change, so I moved the long modification to a lambda

You also had something terrible in the sorting of cards, I even tested it a little and benchmarked it (for comparison):

```
newGetFeatureUsage 184645,277 ops/s
oldGetFeatureUsage     61757,728 ops/s
```

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/CpuManager.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/CpuManager.java
@@ -3,7 +3,6 @@ package me.neznamy.tab.shared;
 import java.util.*;
 import java.util.Map.Entry;
 import java.util.concurrent.*;
-import java.util.concurrent.atomic.LongAdder;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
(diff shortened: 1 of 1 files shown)


NEZNAMY:
This is only used in /tab cpu, which is something most people never use. Performance of the command is not really concerning. Rather than usage being calculated 3x faster, I believe people would prefer to see correct information.
image

zelvalea:
> Это используется только в /tab cpu, который большинство людей никогда не используют. Выполнение команды не очень беспокоит. Вместо того, чтобы использовать в 3 раза быстрее, я считаю, что люди предпочли бы видеть правильную информацию. образ

fixed

zelvalea:
> As you can see, usage is shown 10x lower, which is definitely wrong.

I'm sorry, I'll try and fix it myself

zelvalea:
> Previous output: image Your jar: image
> 
> As you can see, usage is shown 10x lower, which is definitely wrong.

I made a stupid mistake with the merge method ( time + time ), now everything works as expected.

Also added volatile to maintain HB


My build:
(I really have Blackhole.CPU on %group%)
image

BomBardyGamer (approved):
LGTM

PR #933, opened 2023-06-16
Fix loading on velocity

Golfing7:
On velocity, loading the 'IChatBaseComponent' class causes a NoClassDefFoundError. The plugin fails to load because of this.

Error: https://pastebin.com/N39GFLpj


NEZNAMY:
Crazy, I swear it worked when I tested it before. I would say reverting https://github.com/NEZNAMY/TAB/commit/ffbd4bf26b6fa1b7a3bd3a148120ef7510a1c47e will be a better idea to completely remove the code from shared. But how does the same thing for adventure not throw on BungeeCord where it's missing for example?

PR #962, opened 2023-07-04
Fix ViaVersion bug returning -1 version

MachineBreaker:
Explanation:

Both TAB and ViaVersion PlayerJoinEvent priority are LOWEST and seems that TAB registers earlier than ViaVersion the listeners, this causes TAB not being able to get the client protocol version due to ViaVersion not being able to register it firstly.

There might be another way to workaround this since i don't know changing the priority would cause any issue with TAB itself.


c0dingnoobi:
can confirm, pr fixes issue on 1.8 servers (tested using vanilla spigot)

PR #1084, opened 2023-09-23
Fix Arclight compatibility

GsTio86:
This issue has been reported to Arclight, but the developer has not responded yet
I tried to find the problem and successfully fixed it

This problem arises from the existence of two void markDirty methods in Arclight
causes an exception to be thrown

https://github.com/IzzelAliz/Arclight/blob/Trials/arclight-common/src/main/java/io/izzel/arclight/common/mixin/core/network/SynchedEntityDataMixin.java#L49

https://github.com/IzzelAliz/Arclight/issues/1087

### DIFF
--- a/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/nms/DataWatcher.java
+++ b/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/nms/DataWatcher.java
@@ -107,7 +107,11 @@ public class DataWatcher implements EntityData {
             }
         }
         if (BukkitReflection.is1_19_3Plus()) {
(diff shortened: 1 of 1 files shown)


Tanguygab:
instead of a try catch, just use getMethods and get index 0 instead of using getOnlyMethod https://github.com/NEZNAMY/TAB/blob/master/shared/src/main/java/me/neznamy/tab/shared/util/ReflectionUtils.java#L89

GsTio86:
> instead of a try catch, just use getMethods and get index 0 instead of using getOnlyMethod https://github.com/NEZNAMY/TAB/blob/master/shared/src/main/java/me/neznamy/tab/shared/util/ReflectionUtils.java#L89
> 
> or even getMethod, I don't know if there's any difference

Maybe this would be better?

Tanguygab:
perfect :ok_hand:

PR #1107, opened 2023-10-17
Update BukkitTabPlayer.java

ItsVaidas:
Check for permission ignoring OP. I hope this fix will fix the sorting for OP players in the TAB. If not, maybe you yourself can find the exact location and use new Permission(string, PermissionDefault.FALSE);
Needs testing before merging!

### DIFF
--- a/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/BukkitTabPlayer.java
+++ b/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/BukkitTabPlayer.java
@@ -23,6 +23,8 @@ import me.neznamy.tab.shared.backend.BackendTabPlayer;
 import org.bukkit.entity.Player;
 import org.bukkit.metadata.MetadataValue;
 import org.bukkit.potion.PotionEffectType;
+import org.bukkit.permissions.Permission;
+import org.bukkit.permissions.PermissionDefault;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


NEZNAMY:
This could be a breaking change however. People often expect to have full access with OP, which would suddenly break. Even I have no permission system on my test server and just OP myself. I don't know about other people, but is definitely going to have complaints.

ItsVaidas:
Then maybe an option in the config? Because currently my operator is showing above me (owner) and it's getting annoying.

NEZNAMY:
You can for example use groups or configure permissions properly. You can also use your own fork instead of the official release. 

Keep in mind this is a public plugin and according to bStats, it is being used by 26000 servers. Any changes take responsibility. Do you really think your little problem that can easily be solved in many different ways should be solved in a way that may cause confusion for thousands? Are you willing to take the responsibility for that?

ItsVaidas:
No, it's just a suggestion.

I have 2 permission groups and 2 ranks (Which are also permission groups, but primary):
Permission groups: Owner, Operator
Ranks: 1, 2

A player will always have one of two ranks, but permission group is optional.
I can't see a way to sort by a secondary group (which is optional), and primary rank is always excluded to rank.

I don't give everyone OP, just myself and one operator, which helps me operate the server. It is set up correctly, and there is no way to make it any other way with wanted functionality.
I'm thinking about placeholder that would give a letter based on the permission group and grouping by them, but it would be great to have a native support for just ignoring OP permissions in tab sorting

NEZNAMY:
And we're back to what **you** want. This is a public plugin, your change will affect many. You cannot make selfish requests and hope they will be accepted and the needs of everyone else will be put aside. This is what private development is about. You are on the wrong place.

PR #1136, opened 2023-11-09
Added support for PremiumVanishAPI

MrRockis:
I have added the support for PremiumVanish in bungeecord and velocity. Players do not need to have the "tab.seevanished" permission to see people in tablist. **Permission will still override and show everyone.**

PremiumVanish allows groups to have different "vanish levels". This modification uses its api to check if a specific player can see another one. For example you could have admins, mod and builders with different vanish levels. Admins will see everyone, mods only other mods and builder, and builders only other builders.

As premiumvanish is the leading vanish plugin for proxies, I think that this modification is very well needed.

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java
@@ -125,7 +125,7 @@ public class NameTag extends TabFeature implements NameTagManager, JoinListener,
         if (!disableChecker.isDisabledPlayer(disconnectedPlayer) && !hasTeamHandlingPaused(disconnectedPlayer)) {
             for (TabPlayer viewer : TAB.getInstance().getOnlinePlayers()) {
                 if (viewer == disconnectedPlayer) continue; //player who just disconnected
-                if (disconnectedPlayer.isVanished() && !viewer.hasPermission(TabConstants.Permission.SEE_VANISHED) && (premiumVanish == null || !premiumVanish.canSee(viewer.getPlayer(), disconnectedPlayer.getPlayer()))) continue;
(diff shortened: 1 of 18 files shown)


NEZNAMY:
Why does the `canSee` method accept Objects as players and then you check instance and potentially send an error message? You can just use TabPlayer and cast it to BungeeTabPlayer or VelocityTabPlayer for each implementation an then use getPlayer(), which will return platform's player object automatically without the need for casting from object or checking anything.

MrRockis:
I changed the Objects to TabPlayer and also implemented the api for Bukkit.

shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java
```diff
@@ -354,8 +357,7 @@ public void onLoginPacket(TabPlayer player) {
     @Override
     public void onVanishStatusChange(@NotNull TabPlayer player) {
         for (TabPlayer viewer : TAB.getInstance().getOnlinePlayers()) {
-            if (viewer.hasPermission(TabConstants.Permission.SEE_VANISHED)) continue;
-            if (player.isVanished()) {
+            if (player.isVanished() && (!viewer.hasPermission(TabConstants.Permission.SEE_VANISHED) && (premiumVanish == null || !premiumVanish.canSee(viewer, player)))) {
```
NEZNAMY, line 360: Are you sure this will work properly? If a player vanishes, the the will double register their team for players who can see vanished players.

NEZNAMY:
I have made an alternate implementation, please check it out and let me know your thoughts (#1137).

MrRockis:
I think that would work. I wasn't fully sure if you wanted that to be added as a feature or not, but as I saw redisbungee as a feature I used that! :)

NEZNAMY:
You are not using a single advantage of the Feature interfaces, so it's pointless and a method is enough.

MrRockis:
Sounds good! I can try out the implementation after its ready.

NEZNAMY:
It is ready.

PR #1137, opened 2023-11-09
Alternate implementation of #1136

NEZNAMY:


MrRockis:
Seems like everything is **not** working correctly with this build. When you go to vanish your own nametag goes white. Because of this the glow color that you can see also changes to white. This issue was fixed in my commits so there is something wrong.

NEZNAMY:
Do you have the seevanished permission?

MrRockis:
I will need to try this new commit. I'm trying to have it so that I do not need to give the permission to anybody and it would fully work of the premiumvanish api.

MrRockis:
It is fixed that they see their own nametag as white, but there seems to be another issue. If a player is vanished and someone can see them without the permission they have white nametag.

For example is a builder on server is vanished (with vanish level 1) and a moderator (with vanish level 2) comes they just see them with a white nametag while vanished. PremiumVanish API return canSee true. They work correctly in tablist (moderator can see builder).

NEZNAMY:
Yes, see https://github.com/NEZNAMY/TAB/blob/master/shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java#L357-L362

Your implementation caused the plugin to double-register teams, which is an invalid action that used to crash the MC client in the past and still does crash BungeeCord.

MrRockis:
How could this be fixed for everyone? With a quick testing it seems like this is the only problem with this.

NEZNAMY:
The comment says you'd need full tracking of visibility for other players to properly decide on register/unregister actions. Like I said, you can remove the method or use your version that performs an invalid action and crashes BungeeCord, but I don't want to do such thing.

MrRockis:
Are you able to make it so it tracks the visibility of other players? Were you thing about adding a list to each TabPlayer, or what would be the best way?

NEZNAMY:
Yes, something like that would be needed.

NEZNAMY:
I made some simple tracking, but didn't test it. Replace NameTag class with [this](https://pastebin.com/GejrHzGi) and let me know if it works properly.

MrRockis:
We tried a few different ranks and with those test it looked like everything worked **perfectly**! The first version had an issue, but that was fixed with the edit. I'll let you know if we find more issues with this! :)

PR #1202, opened 2024-01-09
Support to use Global Playerlist `server-groups` as Header/Footer `per-server` keys.

Jake-Moore:
Adds support for using Global Playerlist `server-groups` config keys in the Header/Footer `per-server` section.
This allows for configuration of a server group with several velocity/bungee servers which can have a matching config key in the per-server header/footer section.

Example:

```yml
header-footer:
  per-server:
    hubs:
      header: [] #Tab Details Here
      footer: [] #Tab Details Here

global-playerlist:
  server-groups:
    hubs:
    - hub1 # Bungee/Velocity server name
    - hub2 # Bungee/Velocity server name
```

This makes config a little easier and more intuitive because adding another server with a custom TAB is as simple as adding to the global playerlist `server-groups` group, which is then used by header/footer per-server.

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/config/Configs.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/config/Configs.java
@@ -138,21 +138,27 @@ public class Configs {
     }
 
     public String getServerGroup(@NotNull List<Object> serverGroups, @Nullable String server) {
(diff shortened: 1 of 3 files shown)


NEZNAMY:
Since you edited YamlPropertyConfigurationFile file, this will also work for tab/tag stuff. Just wanted to point it out.

PR #1219, opened 2024-01-23
cleaned up the code, added a new method to api

sunmisc:
Cleaned up the code a little, got rid of unnecessary calls to HashMap, changed implementations a little, etc.

What really matters is the API
Passing an array to users is a very bad idea
(getOnlinePlayers())

Firstly, if we really want to transfer an array, then we will always have to copy it, because at any moment it may turn out that someone has changed the array that we are transferring. This is a very bad practice, while we have total Copy On Write, and also Copy On Read

Secondly, it is advisable to get rid of all null returns and use, for example, java.util.Optional, but this is if you are not afraid about backward compatibility, and you haven’t touched it, changed or deleted anything, except that you marked getOnlinePlayers as " Deprecated"

Also, it is advisable to work completely with the api, and not with the implementation, but this is the future, the future

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/TAB.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/TAB.java
@@ -49,11 +49,6 @@ public class TAB extends TabAPI {
     /** Players by their TabList UUID for faster lookup */
     private final Map<UUID, TabPlayer> playersByTabListId = new ConcurrentHashMap<>();
 
-    /** Online player array to avoid memory allocation when iterating */
-
-    @SuppressWarnings("forRemoval")
-    private volatile TabPlayer[] onlinePlayers = new TabPlayer[0];
-
     /** Instance of plugin's main command */
     private TabCommand command;
 
@@ -182,7 +177,7 @@ public class TAB extends TabAPI {
             platform.loadPlayers();
             command = new TabCommand();
             featureManager.load();
-            for (TabPlayer p : onlinePlayers) p.markAsLoaded(false);
+            data.values().forEach(p -> p.markAsLoaded(false));
             if (eventBus != null) eventBus.fire(TabLoadEventImpl.getInstance());
             pluginDisabled = false;
             cpu.enable();
@@ -226,7 +221,6 @@ public class TAB extends TabAPI {
         pluginDisabled = true;
         data.clear();
         playersByTabListId.clear();
-        onlinePlayers = new TabPlayer[0];
         cpu.cancelAllTasks();
     }
 
@@ -239,7 +233,6 @@ public class TAB extends TabAPI {
     public void addPlayer(@NotNull TabPlayer player) {
         data.put(player.getUniqueId(), player);
         playersByTabListId.put(player.getTablistId(), player);
-        onlinePlayers = data.values().toArray(new TabPlayer[0]);
     }
 
     /**
@@ -251,7 +244,6 @@ public class TAB extends TabAPI {
     public void removePlayer(@NotNull TabPlayer player) {
         data.remove(player.getUniqueId());
         playersByTabListId.remove(player.getTablistId());
-        onlinePlayers = data.values().toArray(new TabPlayer[0]);
     }
 
     /**
@@ -289,7 +281,13 @@ public class TAB extends TabAPI {
     }
 
     @Override
-    public @NotNull Collection<me.neznamy.tab.api.TabPlayer> getOnlineTabPlayers() {
+    @Deprecated
+    public @NotNull TabPlayer[] getOnlinePlayers() {
+        return data.values().toArray(new TabPlayer[0]);
+    }
+
+    @Override
+    public @NotNull Collection<TabPlayer> getOnlineTabPlayers() {
         return Collections.unmodifiableCollection(data.values());
     }
 
--- a/shared/src/main/java/me/neznamy/tab/shared/TAB.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/TAB.java
@@ -27,6 +27,8 @@ import org.jetbrains.annotations.Nullable;
 import org.yaml.snakeyaml.error.YAMLException;
 
 import java.io.File;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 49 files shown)


shared/src/main/java/me/neznamy/tab/shared/TAB.java
```diff
@@ -48,6 +50,8 @@ public class TAB extends TabAPI {
     private final Map<UUID, TabPlayer> playersByTabListId = new ConcurrentHashMap<>();
 
     /** Online player array to avoid memory allocation when iterating */
+
+    @SuppressWarnings("forRemoval")
```
NEZNAMY, line 54: Array is used internally to iterate over to avoid creating new iterator each time.

shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardManagerImpl.java
```diff
@@ -35,15 +35,14 @@ public class ScoreboardManagerImpl extends TabFeature implements ScoreboardManag
 
     //defined scoreboards
     @Getter private final Map<String, me.neznamy.tab.api.scoreboard.Scoreboard> registeredScoreboards = new LinkedHashMap<>();
-    private me.neznamy.tab.api.scoreboard.Scoreboard[] definedScoreboards;
```
NEZNAMY, line 38: Same as onlinePlayers array.

shared/src/main/java/me/neznamy/tab/shared/proxy/ProxyTabPlayer.java
```diff
@@ -141,7 +141,7 @@ public boolean hasInvisibilityPotion() {
     public boolean hasPermission(@NotNull String permission) {
         if (TAB.getInstance().getConfiguration().isBukkitPermissions()) {
             sendPluginMessage(new PermissionRequest(permission));
-            return permissions != null && permissions.getOrDefault(permission, false);
+            return permissions.getOrDefault(permission, false);
```
NEZNAMY, line 144: This will throw NPE if assigning groups by permissions & bukkit permission manager is enabled, because permissions will be checked in TabPlayer constructor, before this map was assigned in ProxyTabPlayer constructor.

NEZNAMY:
Passing array is something that can be fixed by returning a copy.
I don't want to switch to optionals, the main reason being it will break the API once again and will provide no advantage.
Working with the API only internally isn't exactly possible, because that would require exposing basically all methods in the API, which would only confuse users and cause them to use internal methods, which will break things.

shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardManagerImpl.java
sunmisc: that's right, I don't see the point in copying the array for any modification (toArray() copies the array)
forEach, in turn, does not copy anything

shared/src/main/java/me/neznamy/tab/shared/TAB.java
sunmisc: the iterator is very cheap
wherein:
getOnlinePlayers() should copy the array
therefore, there is no point in copying the array for modification and storing it; in this case, it makes sense to immediately copy it to getOnlinePlayers()

An iterator costs nothing compared to copying an array (O(n))

shared/src/main/java/me/neznamy/tab/shared/proxy/ProxyTabPlayer.java
sunmisc: most likely I was wrong

sunmisc:
> Passing array is something that can be fixed by returning a copy.
> 
> I don't want to switch to optionals, the main reason being it will break the API once again and will provide no advantage.
> 
> Working with the API only internally isn't exactly possible, because that would require exposing basically all methods in the API, which would only confuse users and cause them to use internal methods, which will break things.
> 
> The ArrayList to HashSet changes are fine assuming they are still working properly.
> 
> However the more complex things such as streams and conditional removal don't really provide any advantages besides increased CPU usage and memory allocations. The plugin already has these values very high when configured to spam RGB animations or using a CPU killing placeholder, which is why I am trying to have internal code as optimized as possible to compensate for it.

For the existing API, alas, it will no longer work well

getOnlinePlayers() must copy the original array for each call
or, as I did, mark the method as deprecated (and it should copy) and create a new method that will return an immutable collection to avoid copying

shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardManagerImpl.java
NEZNAMY: forEach does create an iterator though.

NEZNAMY:
The method is used internally where I am obviously not modifying the array, the only case would be if used by someone else in the API, but that would be shooting yourself in the foot. We can return something else for API, but keep array for internal use. 
The extra memory of the object to due array is none, since it's just another reference and I believe pointers are only 4 bytes (for each player).

sunmisc:
> The method is used internally where I am obviously not modifying the array, the only case would be if used by someone else in the API, but that would be shooting yourself in the foot. We can return something else for API, but keep array for internal use. The extra memory of the object to due array is none, since it's just another reference and I believe pointers are only 4 bytes (for each player).

Yes, exactly, we are talking about the API, and the user who can shoot himself in the foot

But the very fact that this is acceptable is not very good
Whatever one may say, copying an array for modification is an extra operation + the volatile field gives overhead
but you can’t make it non-volatile!

the collection will still be the best solution for a number of reasons:

1) Iteration is comparable in performance to an array
2) The collection can retain some properties for more efficient work with Stream api
3) The API itself now prevents everyone from being shot in the foot

NEZNAMY:
AdventureBossBar class got a modification that's really hard to read for me, was it really necessary to do that? Other than that, I cannot merge this PR due to conflicts.

sunmisc:
> AdventureBossBar class got a modification that's really hard to read for me, was it really necessary to do that? Other than that, I cannot merge this PR due to conflicts.

I'm sorry, give me time and I'll fix some problems, test first

PR #1224, opened 2024-02-02
Fix SpongeForge startup crash

XXMA16:
SpongeForge crashes with the following message `File /path/to/plugin/TAB.v4.1.2.jar constructed 0 mods: [], but had 1 mods specified: [tab]`

I have tested the following versions:
- 1.16.5-36.2.5-8.2.1-RC1459
- 1.16.5-36.2.5-8.2.0

This can be fixed by putting an `@Plugin("tab")` annotation on the main Sponge 8 class.


NEZNAMY:
Is this something specific to SpongeForge? Because vanilla works fine.

sponge8/src/main/java/me/neznamy/tab/platforms/sponge8/Sponge8TAB.java
```diff
@@ -15,13 +15,15 @@
...
 import org.spongepowered.plugin.PluginContainer;
+import org.spongepowered.plugin.builtin.jvm.Plugin;
 
 import java.nio.file.Path;
 
 /**
  * Main class for Sponge 8.
  */
 @Getter
+@Plugin("tab")
```
NEZNAMY, line 26: Can use `TabConstants.PLUGIN_ID`

PR #1335, opened 2024-07-16
Add a signed_texture skin parameter to avoid an API call

qdpy:
I'm adding a `signed_texture` parameter in the skin retrieval process, **which avoids an API call**. For this, I use `signed_texture:texture;signature` in the skin retrieval.


NEZNAMY:
Technically you could just add 2-arg support to `texture`.

PR #1348, opened 2024-08-07
[Layout] prevent layout switching from causing tablist to flicker on simple changes

U5B:
Display name and latency changes can be updated without flickering.
Skins changes still need to be removed/added, which causes flickering.

- ~~ParentGroup is not tested very well, there may be some edgecases.~~

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/features/layout/LayoutManagerImpl.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/layout/LayoutManagerImpl.java
@@ -143,12 +143,12 @@ public class LayoutManagerImpl extends RefreshableFeature implements LayoutManag
         LayoutView current = p.layoutData.view;
         String currentName = current == null ? null : current.getPattern().getName();
         if (!Objects.equals(highestName, currentName)) {
-            // if (current != null) current.destroy();
             if (highest != null) {
                 LayoutView view = new LayoutView(this, highest, p);
                 view.send(current);
                 p.layoutData.view = view;
             } else {
+                if (current != null) current.destroy();
                 p.layoutData.view = null;
             }
         }
--- a/shared/src/main/java/me/neznamy/tab/shared/features/layout/LayoutManagerImpl.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/layout/LayoutManagerImpl.java
@@ -143,12 +143,13 @@ public class LayoutManagerImpl extends RefreshableFeature implements LayoutManag
         LayoutView current = p.layoutData.view;
         String currentName = current == null ? null : current.getPattern().getName();
         if (!Objects.equals(highestName, currentName)) {
-            if (current != null) current.destroy();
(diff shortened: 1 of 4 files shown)


U5B:
I have been working on an addition to Layout that allows you to hide real players on 1.19.3+
This also allows us to use a layout size that is less than 80 slots.
https://github.com/U5B/TAB/commits/usb/hide-real-players/
image

shared/src/main/java/me/neznamy/tab/shared/features/layout/LayoutManagerImpl.java
```diff
@@ -143,12 +143,13 @@ public void refresh(@NotNull TabPlayer p, boolean force) {
         LayoutView current = p.layoutData.view;
         String currentName = current == null ? null : current.getPattern().getName();
         if (!Objects.equals(highestName, currentName)) {
-            if (current != null) current.destroy();
-            p.layoutData.view = null;
+            // if (current != null) current.destroy();
```
NEZNAMY, line 146: What if current is now null? Old one will not be removed.

NEZNAMY:
Why is this even needed? Do you spam players with completely different layouts frequently while they hold the tab key? If yes, I can find a solution (that doesn't break anything, see above) myself without you having to struggle with it (since you aren't as familiar with the plugin).

I don't see the point of hiding real players. This was never really an issue (doesn't it remove them from chat complete too?) or even requested and I see no reason for actually adding it in.

U5B:
> Why is this even needed? Do you spam players with completely different layouts frequently while they hold the tab key? If yes, I can find a solution (that doesn't break anything, see above) myself without you having to struggle with it (since you aren't as familiar with the plugin).

Technically yes, I send them a Layout to refresh whenever a player joins/leaves since I am not using TAB players or RedisBungee. I also plan on having custom layouts per player, configurable by the plugin.

I also didn't want to make a feature request, since I thought it would likely be declined since you aren't accepting feature requests on this project.

> I don't see the point of hiding real players. This was never really an issue (doesn't it remove them from chat complete too?) or even requested and I see no reason for actually adding it in.
It doesn't remove them from chat complete. My reasoning for hiding real players was so that Layout could be resizable, instead of hard coded to 80 slots. So you can have a layout that is only 20 slots like the above image.

NEZNAMY:
> I also didn't want to make a feature request, since I thought it would likely be declined since you aren't accepting feature requests on this project.

That would be an optimization request, those are still accepted.

About the dynamic slot count: This is a 1.19.3+ only solution that would most likely require further work, which counts as a feature request and will not be done for now.

U5B:
> About the dynamic slot count: This is a 1.19.3+ only solution that would most likely require further work, which counts as a feature request and will not be done for now.

If I made a PR for this, would you consider it?

NEZNAMY:
Unfortunately no. It goes further than initial implementation, such as bug fixing, documentation and most importantly customer support. I am trying to reduce the need for people to ask for help, not increase it.

PR #1358, opened 2024-08-24
feat: sayanvanish support

Syrent:
Some users of the SayanVanish plugin have reported bugs related to how TAB handles vanished players. While TAB does remove vanished users from the tab list even without this PR, I believe adding direct support for SayanVanish will help reduce unexpected behaviors.

With this PR, TAB still doesn't remove vanished users if the `layout` mode is enabled or if TAB is installed on Velocity with a shared tab list between multiple servers, and the vanished player is on a different server than the viewer. I'm uncertain if the SayanVanish hook is implemented correctly or if TAB inherently doesn't support vanished users under these specific conditions.

### DIFF
--- a/bukkit/build.gradle.kts
+++ b/bukkit/build.gradle.kts
@@ -15,4 +15,5 @@ dependencies {
         exclude("net.md-5", "bungeecord-chat")
     }
     compileOnly("com.github.LeonMangler:PremiumVanishAPI:2.8.8")
+    compileOnly("org.sayandev:sayanvanish-bukkit:1.5.0-SNAPSHOT")
 }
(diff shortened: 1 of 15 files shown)


Tanguygab:
If vanish status isn't detected with TAB on Velocity, that likely means TAB-Bridge isn't installed on the backend server to check for the vanished metadata

Syrent:
Yes, that seems to have fixed the issue for me. Is there any advantage to adding internal SayanVanish support to TAB?

Tanguygab:
If SayanVanish works on proxy then that's not a bad idea yeah, I don't see any issues with adding it, waiting for NEZ to approve

NEZNAMY:
There is no advantage. Adding this would require maintaining the code and having to update it if there is a breaking change. Adding another repository increases chance of plugin being temporary not compilable if it randomly goes down. The reason for a direct PV hook was due to its vanish levels (so higher level staffs can spy on lower level staffs). If this plugin doesn't have that, there is absolutely no reason to add this. Vanished metadata check exists already, works, and is even more universal and immune to random issues.

Syrent:
I completely understand the concerns with adding something new to maintain, along with the potential issues that can come with it. Introducing another dependency and repository adds risk, as the API could change at any time, or the repository might not be maintained in the long run. Although SayanVanish also has a vanish level system, I agree that it may not be worth the trouble. I’ll consider a more maintainable solution, such as using player metadata for vanish levels or a similar approach. For now, I don't think the vanish level is a critical issue.

PR #1388, opened 2024-11-24
fix: avoid using TAB's internal vanish checks when a registered vanish integration is available

Syrent:
This update ensures that TAB does not check the "vanished" meta if a vanish integration is registered. The redundant check was causing issues with the `canSee` functionality in specific scenarios:  

- A player has the "vanished" meta set (i.e., `isVanished` is true).  
- The registered vanish integration correctly returns `true` for `canSee`.  
- Because `canSee` is true inside integration, TAB continues to check the "vanished" meta. Since the meta indicates the player is vanished, TAB incorrectly overrides the integration’s `canSee` result and returns `false`. 

By skipping the "vanished" meta check when a registered vanish integration is present, this fix ensures the `canSee` behavior aligns with the integration’s logic.

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/proxy/ProxyTabPlayer.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/proxy/ProxyTabPlayer.java
@@ -169,8 +169,11 @@ public abstract class ProxyTabPlayer extends TabPlayer {
 
     @Override
     public boolean isVanished() {
-        for (VanishIntegration i : VanishIntegration.getHandlers()) {
-            if (i.isVanished(this)) return true;
+        if (!VanishIntegration.getHandlers().isEmpty()) {
+            for (VanishIntegration integration : VanishIntegration.getHandlers()) {
+                if (integration.isVanished(this)) return true;
+            }
+            return false;
         }
         return vanished;
     }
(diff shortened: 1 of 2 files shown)


shared/src/main/java/me/neznamy/tab/shared/proxy/ProxyTabPlayer.java
```diff
@@ -169,8 +169,11 @@ public void sendPluginMessage(@NotNull OutgoingMessage message) {
 
     @Override
     public boolean isVanished() {
-        for (VanishIntegration i : VanishIntegration.getHandlers()) {
-            if (i.isVanished(this)) return true;
+        if (!VanishIntegration.getHandlers().isEmpty()) {
```
Syrent, line 172: Checking the same condition in `isVanished` is unnecessary, and I can't think of a valid reason why an integration should return `false` if the player is actually vanished. However, I believe it's better to respect the integration's choice in this case

Syrent:
Everything seems to be working as expected. I’ll mark the PR as ready for review. Additionally, please update the TAB-API to ensure the new VanishIntegration API is accessible through the main API repository.

NEZNAMY:
API is updated on every plugin update and will be done this time as well.

PR #1435, opened 2025-02-17
Expand multi-server support

Rubenicos:
## Redis support -> Proxy support
First of all, any redis-related feature name was renamed to "proxy feature".

### Why?
* I was looking on TAB code to find a solution for multi-server support without using TAB on proxies.
* I find out a pretty well-made feature to transfer messages across proxies wasted on RedisBungee messaging.
* I ~~don't want to mantain this feature with future TAB API updates~~ like to share any plugin edit.

### How it works?
```yaml
proxy-support:
  enabled: true
  # Supported types: PLUGIN, REDIS, RABBITMQ
  type: PLUGIN
  plugin:
    # Compatible plugins: RedisBungee
    # If enabled and compatible plugin is found, hook is enabled to work with proxied players
    name: RedisBungee
  redis:
    url: 'redis://:password@localhost:6379/0'
  rabbitmq:
    exchange: 'plugin'
    url: 'amqp://guest:guest@localhost:5672/%2F'

# On backend servers you should set this option with the ID of the server as defined in your proxy config.
server-name: 'server'
```
By default, the functionality is the same as RedisBungee support configuration.
Adding compatibility for Redis and RabbitMQ using [delivery4j](https://github.com/saicone/delivery4j) library.

**On proxy servers**, it works with global player list.
**On backend servers**, it make global player list work.

### For people that still using TAB v4
I made the same changes on this branch: https://github.com/MineLAT/TAB/tree/v4
You should compile it by yourself.

> [!NOTE]  
> I also add [RGB support on pre 1.16 servers](https://github.com/MineLAT/TAB/commit/f0848c5cd32122d0cd16614dbe1364c89e17ee92) using TAB v4 + ViaVersion, tell me if you want that implementation on TAB v5 to make a pull request later.

### DIFF
--- a/build-logic/src/main/kotlin/tab.platform-conventions.gradle.kts
+++ b/build-logic/src/main/kotlin/tab.platform-conventions.gradle.kts
@@ -11,5 +11,12 @@ tasks {
         relocate("net.kyori.event", "me.neznamy.tab.libs.net.kyori.event")
         relocate("me.neznamy.yamlassist", "me.neznamy.tab.libs.me.neznamy.yamlassist")
         relocate("org.yaml.snakeyaml", "me.neznamy.tab.libs.org.yaml.snakeyaml")
+        relocate("redis.clients.jedis", "me.neznamy.tab.libs.redis.clients.jedis")
+        relocate("com.google.gson", "me.neznamy.tab.libs.com.google.gson")
+        relocate("org.apache.commons.pool2", "me.neznamy.tab.libs.org.apache.commons.pool2")
+        relocate("org.json", "me.neznamy.tab.libs.org.json")
+        relocate("org.slf4j", "me.neznamy.tab.libs.org.slf4j")
+        relocate("com.rabbitmq", "me.neznamy.tab.libs.com.rabbitmq")
+        relocate("com.saicone.delivery4j", "me.neznamy.tab.libs.com.saicone.delivery4j")
     }
 }
--- a/build-logic/src/main/kotlin/tab.platform-conventions.gradle.kts
+++ b/build-logic/src/main/kotlin/tab.platform-conventions.gradle.kts
@@ -12,10 +12,8 @@ tasks {
         relocate("me.neznamy.yamlassist", "me.neznamy.tab.libs.me.neznamy.yamlassist")
         relocate("org.yaml.snakeyaml", "me.neznamy.tab.libs.org.yaml.snakeyaml")
         relocate("redis.clients.jedis", "me.neznamy.tab.libs.redis.clients.jedis")
-        relocate("com.google.gson", "me.neznamy.tab.libs.com.google.gson")
         relocate("org.apache.commons.pool2", "me.neznamy.tab.libs.org.apache.commons.pool2")
         relocate("org.json", "me.neznamy.tab.libs.org.json")
-        relocate("org.slf4j", "me.neznamy.tab.libs.org.slf4j")
         relocate("com.rabbitmq", "me.neznamy.tab.libs.com.rabbitmq")
         relocate("com.saicone.delivery4j", "me.neznamy.tab.libs.com.saicone.delivery4j")
     }
--- a/shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
@@ -307,8 +307,10 @@ public class BelowName extends RefreshableFeature implements JoinListener, QuitL
     public void onJoin(@NotNull ProxyPlayer player) {
         if (TAB.getInstance().getPlatform().isProxy()) return;
         if (player.getBelowNameFancy() == null) return; // This proxy player is not loaded yet
+        // Below name is already being processed by connected player
+        if (TAB.getInstance().isPlayerConnected(player.getUniqueId())) return;
         for (TabPlayer viewer : TAB.getInstance().getOnlinePlayers()) {
-            if (viewer.belowNameData.disabled.get() || viewer.getUniqueId().equals(player.getUniqueId())) continue;
+            if (viewer.belowNameData.disabled.get()) continue;
             viewer.getScoreboard().setScore(
                     OBJECTIVE_NAME,
                     player.getNickname(),
--- a/shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
@@ -303,22 +303,6 @@ public class BelowName extends RefreshableFeature implements JoinListener, QuitL
         }
     }
 
-    @Override
-    public void onJoin(@NotNull ProxyPlayer player) {
-        if (TAB.getInstance().getPlatform().isProxy()) return;
-        if (player.getBelowNameFancy() == null) return; // This proxy player is not loaded yet
-        for (TabPlayer viewer : TAB.getInstance().getOnlinePlayers()) {
-            if (viewer.belowNameData.disabled.get()) continue;
-            viewer.getScoreboard().setScore(
-                    OBJECTIVE_NAME,
-                    player.getNickname(),
-                    player.getBelowNameNumber(),
-                    null, // Unused by this objective slot
-                    player.getBelowNameFancy()
-            );
-        }
-    }
-
     @NotNull
     @Override
     public String getFeatureName() {
--- a/shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
@@ -6,8 +6,8 @@ import me.neznamy.tab.shared.TAB;
 import me.neznamy.tab.shared.TabConstants;
 import me.neznamy.tab.shared.cpu.ThreadExecutor;
 import me.neznamy.tab.shared.cpu.TimedCaughtTask;
-import me.neznamy.tab.shared.features.redis.RedisPlayer;
-import me.neznamy.tab.shared.features.redis.RedisSupport;
+import me.neznamy.tab.shared.features.proxy.ProxyPlayer;
+import me.neznamy.tab.shared.features.proxy.ProxySupport;
 import me.neznamy.tab.shared.features.types.*;
 import me.neznamy.tab.shared.placeholders.conditions.Condition;
 import me.neznamy.tab.shared.platform.Scoreboard;
@@ -24,7 +24,7 @@ import java.util.Map;
  * Feature handler for BelowName feature
  */
 public class BelowName extends RefreshableFeature implements JoinListener, QuitListener, Loadable,
-        WorldSwitchListener, ServerSwitchListener, CustomThreaded, RedisFeature, VanishListener {
+        WorldSwitchListener, ServerSwitchListener, CustomThreaded, ProxyFeature, VanishListener {
 
     /** Objective name used by this feature */
     public static final String OBJECTIVE_NAME = "TAB-BelowName";
@@ -44,7 +44,7 @@ public class BelowName extends RefreshableFeature implements JoinListener, QuitL
     private final DisableChecker disableChecker;
 
     @Nullable
-    private final RedisSupport redis = TAB.getInstance().getFeatureManager().getFeature(TabConstants.Feature.REDIS_BUNGEE);
+    private final ProxySupport proxy = TAB.getInstance().getFeatureManager().getFeature(TabConstants.Feature.PROXY_SUPPORT);
 
     /**
      * Constructs new instance and registers disable condition checker and text refresher to feature manager.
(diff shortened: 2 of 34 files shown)


shared/src/main/java/me/neznamy/tab/shared/features/proxy/ProxyMessengerSupport.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.function.Supplier;
+
+public class ProxyMessengerSupport extends ProxySupport {
+
+    private final Supplier<Broker> brokerSupplier;
+    private AbstractMessenger messenger;
+
+    public ProxyMessengerSupport(Supplier<Broker> brokerSupplier) {
+        this.brokerSupplier = brokerSupplier;
```
NEZNAMY, line 16: can use @ RequiredArgsConstructor

shared/src/main/java/me/neznamy/tab/shared/features/playerlist/PlayerList.java
```diff
@@ -276,9 +276,13 @@ public void onJoin(@NotNull TabPlayer connectedPlayer) {
...
             }
-            if (redis != null) {
-                for (RedisPlayer redis : redis.getRedisPlayers().values()) {
-                    connectedPlayer.getTabList().updateDisplayName(redis.getUniqueId(), redis.getTabFormat());
+            if (proxy != null) {
+                for (ProxyPlayer proxied : proxy.getProxyPlayers().values()) {
+                    if (TAB.getInstance().getPlatform().isProxy()) {
+                        connectedPlayer.getTabList().updateDisplayName(proxied.getUniqueId(), proxied.getTabFormat());
+                    } else {
+                        connectedPlayer.getTabList().addEntry(proxied.asEntry());
```
NEZNAMY, line 284: Is this an attempt to make this feature an alternative to global playerlist? What if someone doesn't want this functionality between different servers? Players on the same server should already be added into the tablist by the server itself.

shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
```diff
@@ -292,10 +292,30 @@ public void onQuit(@NotNull TabPlayer disconnectedPlayer) {
...
+        if (player.getBelowNameFancy() == null) return; // This proxy player is not loaded yet
+        for (TabPlayer viewer : TAB.getInstance().getOnlinePlayers()) {
+            if (viewer.belowNameData.disabled.get()) continue;
+            viewer.getScoreboard().setScore(
+                    OBJECTIVE_NAME,
+                    player.getNickname(),
+                    player.getBelowNameNumber(),
+                    null, // Unused by this objective slot
+                    player.getBelowNameFancy()
+            );
```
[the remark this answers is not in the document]
Rubenicos: Maybe, unlike nametag and yellow number, this one doesn't seems to be at the end of the class

shared/src/main/java/me/neznamy/tab/shared/features/playerlist/PlayerList.java
Rubenicos: Actually its a fix, since TAB doesn't show proxied players on backend using a display name update

shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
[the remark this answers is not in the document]
NEZNAMY: The proxy message class is separated in this one, it's not nested. I started refactoring, but didn't finish it.

shared/src/main/java/me/neznamy/tab/shared/features/playerlist/PlayerList.java
NEZNAMY: So backend server won't add the players into the tablist?

NEZNAMY:
Overall interesting. I wouldn't call the potential wasted, as RedisBungee has been requested and used by more servers than this functionality, which has never been requested.

Viaversion hook sounds interesting. I don't know if there are many <1.16 servers that allow 1.16+ and want to show RGB colors to them. But I guess having it shouldn't hurt. I am more interested in NumberFormat support though. You only need to implement null (shouldn't be used by the plugin, but keeping it nullable because it is) and FixedFormat. The remaining two are redundant.

shared/src/main/java/me/neznamy/tab/shared/features/playerlist/PlayerList.java
Rubenicos: It actually add the players, Im using this changes right now in my server.
For example:
* SkyBlock mode is separated in 2 servers: islands & spawn
* Both servers use TAB messenger via redis.
* Both servers show all the connected players.

NEZNAMY: You just described global playerlist feature.

Rubenicos: Yes, but using this changes is not required to have TAB on proxy servers

NEZNAMY: I think it should still be more configurable like global playerlist rather than forced on. You also seem to be sending the packet to the players on the same server too, which isn't ideal. In fact, since 1.19 it will ignore the duplicated entry, not even updating display name correctly.

Rubenicos: Of course, making an implementation of global player list for backend servers should be a better solution, but my knowledge about TAB peaked on this pull request xd
If you have some start point or a configuration example, I whould can help with it.

Also, the players on the same server are not being duplicated since they are not saved as ProxyPlayer on the same backend server.

NEZNAMY: Honestly just enabling global playerlist (when it's enabled) should pretty much do the trick. The default functionality will never go through, since server check will always cancel every operation. However, proxy (redis) operations should be executed as intended.

shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java
```diff
@@ -411,12 +415,13 @@ public void onRedisLoadRequest() {
...
     @Override
-    public void onQuit(@NotNull RedisPlayer player) {
+    public void onQuit(@NotNull ProxyPlayer player) {
         if (player.getTeamName() == null) {
-            TAB.getInstance().getErrorManager().printError("Unable to unregister team of redis player " + player.getName() + " on quit, because team is null", null);
+            TAB.getInstance().getErrorManager().printError("Unable to unregister team of proxy player " + player.getName() + " on quit, because team is null", null);
             return;
         }
         for (TabPlayer viewer : onlinePlayers.getPlayers()) {
+            if (viewer.getUniqueId().equals(player.getUniqueId())) continue;
```
NEZNAMY, line 424: There checks contradict what you said previously (and how the plugin works) - you won't get TabPlayer and ProxyPlayer for the same player.

shared/src/main/java/me/neznamy/tab/shared/features/playerlist/PlayerList.java
Rubenicos: Done, but not yet tested

shared/src/main/java/me/neznamy/tab/shared/features/globalplayerlist/GlobalPlayerList.java
```diff
@@ -291,6 +291,7 @@ public void refresh(@NotNull TabPlayer refreshed, boolean force) {
     private boolean shouldSee(@NotNull TabPlayer viewer, @NotNull ProxyPlayer target) {
         if (target.isVanished() && !viewer.hasPermission(TabConstants.Permission.SEE_VANISHED)) return false;
         if (viewer.globalPlayerListData.onSpyServer) return true;
+        if (viewer.getUniqueId().equals(target.getUniqueId())) return false;
```
NEZNAMY, line 294: As discussed before, there should not be TabPlayer and ProxyPlayer for the same player.

shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java
Rubenicos: But now that I think about it, the check should be different, since the bug will not happen with the player, but with other connected players.

Rubenicos: I will cancel any incoming message related with an already connected player

NEZNAMY: Just keep in mind canceling messages will result in them never being processed, rather than processed later, which sounds like the solution you'd need.

NEZNAMY:
Can you test global playerlist and check out if the new belowname code is actually needed? Then, it might be ready for merge.

build-logic/src/main/kotlin/tab.platform-conventions.gradle.kts
```diff
@@ -11,5 +11,12 @@ tasks {
         relocate("net.kyori.event", "me.neznamy.tab.libs.net.kyori.event")
         relocate("me.neznamy.yamlassist", "me.neznamy.tab.libs.me.neznamy.yamlassist")
         relocate("org.yaml.snakeyaml", "me.neznamy.tab.libs.org.yaml.snakeyaml")
+        relocate("redis.clients.jedis", "me.neznamy.tab.libs.redis.clients.jedis")
+        relocate("com.google.gson", "me.neznamy.tab.libs.com.google.gson")
+        relocate("org.apache.commons.pool2", "me.neznamy.tab.libs.org.apache.commons.pool2")
+        relocate("org.json", "me.neznamy.tab.libs.org.json")
+        relocate("org.slf4j", "me.neznamy.tab.libs.org.slf4j")
+        relocate("com.rabbitmq", "me.neznamy.tab.libs.com.rabbitmq")
+        relocate("com.saicone.delivery4j", "me.neznamy.tab.libs.com.saicone.delivery4j")
```
NEZNAMY, line 20: I didn't realize you are trying to tenfold the jar size. Are these really necessary? Google gson is available on every 1.8+ server, what would you even use slf4j for?

Rubenicos: Redis library depends on `com.google.code.gson:gson`, `org.apache.commons:commons-pool2`, `org.json:json` and `org.slf4j:slf4j-api`.
While RabbitMQ depends on `org.slf4j:slf4j-api`.

Slf4j API requires Slf4j NOP if there is no implementation.

I removed the shading of gson and slf4j, apparently most platforms have those libraries.

Rubenicos:
> Can you test global playerlist and check out if the new belowname code is actually needed? Then, it might be ready for merge.

Technically is not needed since you cannot see a player entity that is not in the same backend server, but I don't know if it's needed for platforms like [MultiPaper](https://github.com/MultiPaper/MultiPaper).

Btw, I will do some testing later, most of the changes made on this pull are not tested yet with TAB v5.

NEZNAMY:
Go ahead with the tests. Check the belowname too if the addition is really needed. What is the current jar size?

Rubenicos:
> Go ahead with the tests. Check the belowname too if the addition is really needed. What is the current jar size?

After some tests, the plugin works surprisingly well (I tried to use global player list with TAB v4 and doesn't work), and belowname with proxy support doesn't seems to be needed at all, should I remove it?

The .jar size is more than 3MB (due Redis and RabbitMQ libraries), if the size is a problem I also have a [library loader](https://github.com/saicone/ezlib).

NEZNAMY:
If the extra belowname code is not needed, remove it obviously. Doubling the jar size for 1 server out of 36000 is quite unfortunate though. Library loading is something I don't want to have, so it's better shaded.

NEZNAMY:
Looks fine to me. Anything else before merging?

Rubenicos:
Anything else, in a few days I will make another pull for ViaVersion hook (including fancy value support on pre 1.20.3 servers)

NEZNAMY:
The build is failing for me, as well as github. How is it that it worked for you?

Rubenicos:
It doesn't xd, apparently is an error with fabric build, that is shading the whole fabric server after append the module code into final jar module.
I thought it was an error from my IDE cache (and disabled fabric module build), but it seems that is an error that appeared since the plugin is now using more libraries.

NEZNAMY:
Yeah the fabric mess is unfortunate. Since I had no experience with it, the module was added by a person who then ditched on me and left me with a module that takes forever to compile and creates massive jars (though the final jar is fine).

Rubenicos:
Someone reported a similar problem on [fabric discord](https://discord.com/channels/507304429255393322/1294856025106808946), apparently can be fixed by remaking the whole build configuration, using the mod template as start, I will see it in a bit.

NEZNAMY:
I have found a fix in the meantime.

PR #1443, opened 2025-03-01
Add ViaVersion packets support

Rubenicos:
> [!TIP]
> Known bugs:
> * ~~Somethings TAB send tablist and scoreboard information too early (before ViaVersion do all the stuff to properly send packets to players) causing to not see any of both, can be fixed for scoreboard if a delay is set, and for tablist only with `/tab reload`.~~

> [!NOTE]  
> Only Bukkit support by default, **I think** this can be used as well on Sponge module.

### Additions
* [Tablist] Hat update for +1.21.4 players on pre 1.21.4 servers.
* [Tablist] Text shadow color support for +1.21.4 players on pre 1.21.4 servers.
* [Tablist] List order support for +1.21.2 players on pre 1.21.2 servers.
* [Tablist] Listed update for +1.19.3 players on pre 1.19.3 servers.
* [Tablist] RGB colors support for +1.16 players on pre 1.16 servers.
* [Tablist] Any other stuff that I didn't mention above for +1.16 players on pre 1.16 servers.
* [Scoreboard] Text shadow color support for +1.21.4 players on pre 1.21.4 servers.
* [Scoreboard] Number format support for +1.20.3 players on pre 1.20.3 servers.
* [Scoreboard] RGB colors support for +1.16 players on pre 1.16 servers.
* [Scoreboard] Any other stuff that I didn't mention above for +1.13 players on pre 1.13 servers.

~~I used a separated module, as "viaversion platform" since it doesn't require more than a player id to send packets.~~


NEZNAMY:
Wouldn't it be better to have this in shared? There already are some existing implementations using adventure api for example.

component/src/main/java/me/neznamy/chat/TextColor.java
```diff
@@ -125,6 +125,15 @@ private EnumChatFormat loadClosestColor() {
         return closestColor;
     }
 
+    /**
+     * Check if the current text color is a legacy color code.
+     *
+     * @return true if this a legacy color, false otherwise.
+     */
+    public boolean isLegacy() {
+        return legacyColor != null;
```
NEZNAMY, line 134: legacyColor not being null can also mean the closest color was calculated on request.

viaversion/src/main/java/me/neznamy/tab/platforms/viaversion/scoreboard/ViaScoreboard.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public void updateObjective(@NonNull Objective objective) {
+        sendObjectiveUpdate(ObjectiveAction.UPDATE, objective);
+    }
+
+    @Override
+    public @NotNull Object createTeam(@NonNull String name) {
+        return null;
```
NEZNAMY, line 82: NotNull

NEZNAMY:
How does this work exactly? You write the packet the way it was seen at some point and via remaps it? That wouldn't explain the need for so many classes though.

Rubenicos:
ViaVersion use "the best" or "the correct way" to remap packets across versions, for example, if your server is on 1.16.5 and a player on 1.21.4 connects, any packet that is sent to player first need to be converted from 1.16.5 -> 1.17.x -> 1.18.x -> ... etc. until it reach to player version, that is the main reason it uses packages like `v1_19_1to1_19_3` for **every** protocol update.

If you create a packet using 1.17 protocol and send it to a 1.21.4 player, ViaVersion will do all the thing to make that packet reach the player.

So, I selected every protocol that makes a change on a packet and make a packet write for its changes, for example, the packet `PLAYER_INFO` is used to send tablist information and its changes are:
* 1.16 - RGB colors.
* 1.19.3 - Separated packet to remove entries, `BitSet` usage for actions and `UPDATE_LISTED` action.
* 1.21.2 - `UPDATE_LIST_ORDER` action.
* 1.21.4 - Shadow color and `UPDATE_HAT` action.

Next, I made an extension from abstract class (that contains the shared logic between version changes), this is because I see some classes on bukkit module named like "PacketTabList1193", but I can made it using nested classes or a protocol version check in the same class as well, tell me what do you think.

The other thing is the fabulous way how ViaVersion work, if ViaVersion gets updated, this changes will still working using the same TAB version, since ViaVersion "stacks" new protocols above old ones (following "the correct way") by adding a new protocol remapper without affecting the protocols that are currently used.

viaversion/src/main/java/me/neznamy/tab/platforms/viaversion/scoreboard/ViaScoreboard.java
NEZNAMY: The issue is that the method is not allowed to return null, not the annotation itself. use `return new Object();`.

NEZNAMY:
So can we not just create a packet for the latest version and let Via remap it back? Or would that also require ViaBackwards which is why you didn't do it?

component/src/main/java/me/neznamy/chat/component/TabComponent.java
```diff
@@ -102,6 +103,11 @@ public abstract class TabComponent {
     @Setter
     private Component adventureComponent;
 
+    /** ViaVersion component from this component */
+    @Nullable
+    @Setter
+    private Object viaComponent;
```
NEZNAMY, line 109: Why is this an object when it's always JsonObject? Same for all related methods.

component/src/main/java/me/neznamy/chat/hook/ViaVersionHook.java
```diff
@@ -0,0 +1,117 @@
...
+            return ComponentUtil.legacyToJson(component.toLegacyText());
+        } else if (component instanceof TextComponent) {
+            return convert((TextComponent) component);
+        } else if (component instanceof TranslatableComponent) {
+            return convert((TranslatableComponent) component);
+        } else if (component instanceof KeybindComponent) {
+            return convert((KeybindComponent) component);
+        } else {
+            throw new IllegalStateException("Unknown component type: " + component.getClass().getName());
+        }
```
NEZNAMY, line 32: Why splitting the methods? Component style and extra are not limited to text component, just insert the key & value based on component type and always insert style.

component/src/main/java/me/neznamy/chat/hook/ViaVersionHook.java
```diff
@@ -0,0 +1,117 @@
...
+        if (modifier.getShadowColor() != null) {
+            object.addProperty("shadow_color", modifier.getShadowColor());
+        }
+
+        // Render
+        if (Boolean.TRUE.equals(modifier.getBold())) {
+            object.addProperty("bold", true);
+        }
+        object.addProperty("italic", Boolean.TRUE.equals(modifier.getItalic()));
```
NEZNAMY, line 61: Is this intentionally different than other magic codes?

Rubenicos:
> So can we not just create a packet for the latest version and let Via remap it back? Or would that also require ViaBackwards which is why you didn't do it?

Indeed, it is preferable to use only ViaVersion logic since this changes aims to achieve new features compatibility with players in newer versions.

component/src/main/java/me/neznamy/chat/component/TabComponent.java
Rubenicos: Is an `Object` due is not a `com.google.gson.JsonObject`, it is a `com.viaversion.viaversion.libs.gson.JsonObject`

NEZNAMY: And what's the problem?

component/src/main/java/me/neznamy/chat/hook/ViaVersionHook.java
```diff
@@ -0,0 +1,117 @@
...
+
+        // Root
+        object.addProperty("type", "text"); // +1.20.3 optimization
+        object.addProperty("text", component.getText());
+
+        // Color
+        final ChatModifier modifier = component.getModifier();
+        if (modifier.getColor() != null) {
+            final TextColor color = modifier.getColor();
+            if (color.isLegacy()) {
```
NEZNAMY, line 47: I'm not 100% sure about this one. It might need a version check. Plugin might try to use a RGB code and then send it to 1.13 player this way.

component/src/main/java/me/neznamy/chat/hook/ViaVersionHook.java
Rubenicos: It's a "bug" that exists since raw json text was created by Mojang, if you don't specify the text is not italic, it will show as italic text by default.

component/src/main/java/me/neznamy/chat/hook/ViaVersionHook.java
Rubenicos: I made a separated component send on ViaScoreboard13 to avoid RGB usage on pre 1.16 players.

NEZNAMY: Then what's the point of this check? If legacy colors are used, they are converted to TextColor, which contains the correct rgb code that can be used.

component/src/main/java/me/neznamy/chat/component/TabComponent.java
Rubenicos: If ViaVersion is not loaded it might fail with `NoClassDefFoundError`, I guess.

viaversion/src/main/java/me/neznamy/tab/platforms/viaversion/platform/ViaVersionPlatform.java
```diff
@@ -0,0 +1,95 @@
...
+import me.neznamy.tab.shared.platform.TabList;
+import me.neznamy.tab.shared.platform.TabPlayer;
+import me.neznamy.tab.shared.util.function.FunctionWithException;
+import org.jetbrains.annotations.NotNull;
+
+public class ViaVersionPlatform {
+
+    @NotNull
+    public static <P extends TabPlayer> FunctionWithException<P, Scoreboard> findScoreboardProvider(@NotNull ProtocolVersion serverVersion) {
+        FunctionWithException<P, Scoreboard> function = FunctionWithException.empty();
```
NEZNAMY, line 21: Can't you just return immediately without assigning to variable? (will require swapping the checks)

component/src/main/java/me/neznamy/chat/component/TabComponent.java
NEZNAMY: It doesn't do that for adventure component though.

component/src/main/java/me/neznamy/chat/hook/ViaVersionHook.java
Rubenicos: It's a client-side micro-optimization to send (for example) `"color": "white"` instead of `"color": "#ffffff"`

NEZNAMY: Realistically, is it? It just adds more code here, which can confuse someone.

viaversion/src/main/java/me/neznamy/tab/platforms/viaversion/platform/ViaVersionPlatform.java
Rubenicos: Do you mean something like this?
```java
if (serverVersion.getNetworkId() < ProtocolVersion.V1_13.getNetworkId()) {
    return player -> {
        if (player.getVersion().getNetworkId() >= ProtocolVersion.V1_20_3.getNetworkId()) {
            return new ViaScoreboard1203<>(player);
        } else if (player.getVersion().getNetworkId() >= ProtocolVersion.V1_16.getNetworkId()) {
            return new ViaScoreboard16<>(player);
        } else if (player.getVersion().getNetworkId() >= ProtocolVersion.V1_13.getNetworkId()) {
            return new ViaScoreboard13<>(player);
        }
        return null;
    };
} else if (serverVersion.getNetworkId() < ProtocolVersion.V1_16.getNetworkId()) {
    return player -> {
        if (player.getVersion().getNetworkId() >= ProtocolVersion.V1_16.getNetworkId()) {
            return new ViaScoreboard16<>(player);
        } else if (player.getVersion().getNetworkId() >= ProtocolVersion.V1_13.getNetworkId()) {
            return new ViaScoreboard13<>(player);
        }
        return null;
    };
} else if (serverVersion.getNetworkId() < ProtocolVersion.V1_20_3.getNetworkId()) {
    return player -> {
        if (player.getVersion().getNetworkId() >= ProtocolVersion.V1_20_3.getNetworkId()) {
            return new ViaScoreboard1203<>(player);
        }
        return null;
    };
}
return FunctionWithException.empty();
```

NEZNAMY: Yes, something like that.

component/src/main/java/me/neznamy/chat/hook/ViaVersionHook.java
```diff
@@ -0,0 +1,113 @@
...
+import me.neznamy.chat.component.TextComponent;
+import me.neznamy.chat.component.TranslatableComponent;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public class ViaVersionHook {
+
+    @NotNull
+    public static Object convert(@NotNull TabComponent component) {
```
NEZNAMY, line 21: You could merge all of these methods into one.

component/src/main/java/me/neznamy/chat/hook/ViaVersionHook.java
NEZNAMY: Actually, there is a difference between false and null. Not specifying (null) instead of false would return in magic codes not being reset when they are supposed to be. Right?

viaversion/src/main/java/me/neznamy/tab/platforms/viaversion/scoreboard/ViaScoreboard1203.java
```diff
@@ -0,0 +1,103 @@
...
+    protected void writeObjective(@NonNull PacketWrapper packet, @NonNull Objective objective) {
+        packet.write(Types.VAR_INT, objective.getDisplaySlot().ordinal());
+        packet.write(Types.STRING, objective.getName());
+    }
+
+    @Override
+    protected void writeObjectiveDisplay(@NonNull PacketWrapper packet, @NonNull Objective objective) {
+        super.writeObjectiveDisplay(packet, objective);
+        // ViaVersion configuration compatibility
+        if (objective.getHealthDisplay() == HealthDisplay.INTEGER && Via.getConfig().hideScoreboardNumbers()) {
```
NEZNAMY, line 79: We don't want to respect this viaversion option. Currently, it causes `score` to not be visible for 1.20.3+ players on <1.20.3 servers, but not only in sidebar, but also belowname and playerlist. This is an opportunity to resolve this problem.

viaversion/src/main/java/me/neznamy/tab/platforms/viaversion/scoreboard/ViaScoreboard.java
NEZNAMY: This isn't resolved.

Rubenicos:
In summary:
* I forgot to send components as NBT for +1.20.3 players, so I created another method on TabComponent (Still need to test if a NoClassDefFoundError occurs or not).
* Added a dummy object for unused scoreboard team.
* Added safe check to get skin data from bukkit player.

Still having some errors due TAB send tablist and scoreboard information too early (before ViaVersion do all the stuff to properly send packets to players) causing to not see any of both somethings, can be fixed for scoreboard if a delay is set, but with tablist I didn't see an delay on default configuration.

viaversion/src/main/java/me/neznamy/tab/platforms/viaversion/scoreboard/ViaScoreboard.java
```diff
@@ -158,9 +158,11 @@ protected void sendTeamUpdate(int mode, @NonNull Team team) {
         packet.scheduleSend(protocol);
     }
 
+    protected abstract void writeComponent(@NonNull PacketWrapper packet, @NonNull String text);
```
NEZNAMY, line 163: Can replace usage with `new SimpleTextComponent(text)` and no need for 2 methods.

viaversion/src/main/java/me/neznamy/tab/platforms/viaversion/tablist/ViaTabList1214.java
```diff
@@ -0,0 +1,81 @@
...
+                break;
+            default:
+                throw new IllegalStateException("Cannot write " + action.name() + " for 1.21.4 packet");
+        }
+    }
+
+    @Override
+    protected void writeEntry(@NonNull PacketWrapper packet, @NonNull Entry entry) {
+        super.writeEntry(packet, entry);
+        packet.write(Types.VAR_INT, entry.getListOrder());
```
NEZNAMY, line 78: 1212 already writes list order.

viaversion/src/main/java/me/neznamy/tab/platforms/viaversion/tablist/ViaTabList16.java
```diff
@@ -0,0 +1,116 @@
...
+            // Properties size
+            packet.write(Types.VAR_INT, 1);
+            // Properties
+            packet.write(Types.STRING, entry.getName());
+            packet.write(Types.STRING, entry.getSkin().getValue());
+            packet.write(Types.OPTIONAL_STRING, entry.getSkin().getSignature());
+        }
+        packet.write(Types.VAR_INT, entry.getGameMode());
+        packet.write(Types.VAR_INT, entry.getLatency());
+        packet.write(Types.OPTIONAL_COMPONENT, entry.getDisplayName() == null ? null : entry.getDisplayName().toViaVersion());
```
NEZNAMY, line 114: can use #writeOptionalComponent

NEZNAMY:
If you move the classes into Bukkit module, you'll be able to use `SkinData` class to properly get player's skin.

NEZNAMY:
1.21.5-pre1 is out. We can expect the full release within the next weeks. Do you think we could get this PR fully resolved by then?

Rubenicos:
Should I move viaversion module into bukkit module maintaining the package `me.neznamy.tab.platforms.viaversion` or move to `me.neznamy.tab.platforms.bukkit`?

NEZNAMY:
Move it to bukkit package.

NEZNAMY:
Maybe moving the classes (scoreboard & tablist) to sub-packages called viaversion would be a good idea to make it not be in the same package as other implementations.

I have also noticed some inconsistency. You call 1.13 and 1.16 classes `13` and `16`, while calling for example 1.20.3 `1203`. Shouldn't they be called `113` and `116` for consistency?

NEZNAMY:
Yes.

NEZNAMY:
Were you able to make any progress regarding the issue on join you described? Perhaps viaversion people might be able to help?

Rubenicos:
Added a first-time 0.5 seconds delay using player channel event loop, after some tests seems to be needed 0.4 seconds or more to avoid any error, so I used 0.5 to make sure it works.

NEZNAMY:
That sounds very unreliable. Is there no way to detect this? Do none of the variables inside viaversion change?

bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/tablist/viaversion/ViaTabList1193.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    public void removeEntry(@NonNull UUID entry) {
+        final PacketWrapper packet = PacketWrapper.create(playerInfoRemove, null, connection);
+
+        // Players
+        packet.write(Types.UUID_ARRAY, new UUID[] { entry });
+
+        packet.scheduleSend(protocol);
```
NEZNAMY, line 54: Forgot to call delayed send here?

Rubenicos:
> That sounds very unreliable. Is there no way to detect this? Do none of the variables inside viaversion change?

I checked the pipeline, and everything seems to be fine (ViaVersion is already registered when ViaTabList and ViaScoreboard constructor is called), so I think this is a netty-related behaviour.

> Also this may result in packets being sent out of order when sent just after the delay.

* The player joins
* TAB send multiple packets in the same thread
* ViaTabList / ViaScoreboard make a delayed execution for those packets (since they are in the same thread) using `PacketWrapper#send()` to send them immediately after delay
* After delay any packet is sent using `PacketWrapper#scheduleSend()` in the same event loop that is used for delayed executions, so technically they will not interfere with packets that was sent before

NEZNAMY:
So let's say at 400ms you send a packet. There will be a 500ms delay, resulting in it being sent at 900ms. What if a packet is sent at 600ms? Will it be sent after or before?

Rubenicos:
Using a `Queue` is more consistent, if I could find a way to detect that when the channel context will correctly send the packet I could make a better solution

NEZNAMY:
Everything looks fine to me now. Have all server-client version pairs been tested?

Rubenicos:
Tested by:
* Joining on 1.8 server using 1.21.4, 1.21.2, 1.19.4 and 1.16.5
* Joining on 1.12 server using 1.21.4, 1.21.2, 1.19.4 and 1.16.5
* Joining on 1.17.1 server using 1.21.4, 1.21.2 and 1.19.4
* Joining on 1.20.1 server using 1.21.4, and 1.21.2

Everything using default TAB configuration + enabling scoreboard

NEZNAMY:
Have you tried 1.13 client too? For scoreboard. Otherwise, sounds good.

Rubenicos:
Now tested by:
* Joining on 1.8 server using 1.13.2
* Joining on 1.12.2 server using 1.13.2

NEZNAMY:
Anything else before merging?

Rubenicos:
Anything else.

Btw, I known unlimited nametag mode was removed since v5.
My server currently use an additional plugin for that, but I think this can be achieved in a cleaner way by providing an empty API inside TAB code.
For example, if TAB doens't found a UnlimitedNametag provider (that will never find inside the TAB code) it will work as normal.

NEZNAMY:
What's the point of adding an empty interface that will never be instantiated?

Rubenicos:
To allow other people to register its own implementation, for example, an implementation compatible with ModelEngine that shows models as name decorations

NEZNAMY:
I think that is quite of out of scope, especially for a feature that was literally removed. I have no interest in doing anything related to it.

Rubenicos:
Understandable

NEZNAMY:
Thank you for your contribution.

NEZNAMY:
I have seen someone report the delay sometimes not being enough. After printing objects of the connection before and after the delay, I noticed one difference:

```
protocolInfo: ProtocolInfo{clientState=LOGIN, serverState=PLAY, protocolVersion=1.21.4 (769), serverProtocolVersion=1.8.x (47), username='_NEZNAMY_', uuid=6603bf58-226d-3b2b-8f5e-2f5d89104533}

protocolInfo: ProtocolInfo{clientState=PLAY, serverState=PLAY, protocolVersion=1.21.4 (769), serverProtocolVersion=1.8.x (47), username='_NEZNAMY_', uuid=6603bf58-226d-3b2b-8f5e-2f5d89104533}
```

Perhaps the reason why it didn't work is because those packets cannot be sent during login/configuration state, which for some reason lasts way longer than it should. Would you have any idea how to solve this with some kind of event listening?

Rubenicos:
Brilliant!
I now know how to send the packets properly

I'll make a pull request in a bit

bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/tablist/viaversion/ViaTabList116.java
```diff
@@ -0,0 +1,114 @@
...
+    protected void writeEntry(@NonNull PacketWrapper packet, @NonNull Entry entry) {
+        packet.write(Types.STRING, entry.getName());
+        if (entry.getSkin() == null) {
+            // Properties size
+            packet.write(Types.VAR_INT, 0);
+        } else {
+            // Properties size
+            packet.write(Types.VAR_INT, 1);
+            // Properties
+            packet.write(Types.STRING, entry.getName());
```
NEZNAMY, line 106: This was supposed to be "textures", not player name.

bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/tablist/viaversion/ViaTabList1193.java
```diff
@@ -0,0 +1,105 @@
...
+    protected void writeEntry(@NonNull PacketWrapper packet, @NonNull Entry entry) {
+        packet.write(Types.STRING, entry.getName());
+        if (entry.getSkin() == null) {
+            // Properties size
+            packet.write(Types.VAR_INT, 0);
+        } else {
+            // Properties size
+            packet.write(Types.VAR_INT, 1);
+            // Properties
+            packet.write(Types.STRING, entry.getName());
```
NEZNAMY, line 95: Same here.

NEZNAMY:
While this hook is cool, I am not sure about keeping it up for the future. Some of the current concerns are:
* Complicated code to create and maintain (see above for a random issue someone found which was caused by a mistake).
* Increased risk of causing packet encoding errors for users, and I definitely don't want to bring in instability.
* Starting with ViaVersion build 1144, tablist usage now throws packet rewriting errors. The more I get into investigation, the more confused I am. 1.21.5 gets `[ViaVersion] ERROR IN Protocol1_21_4To1_21_5 IN REMAP OF PLAYER_INFO_UPDATE (0x40)`. The final stage to where I got was only keeping 1.16 implementation and sending it on 1.8 server. Result? 1.16.5 player can see everything correctly, but 1.21.x gets `[ViaVersion] ERROR IN Protocol1_18_2To1_19 IN REMAP OF PLAYER_INFO (0x36)`. Everything worked fine in build 1143. I can't even report this to ViaVersion issue tracker, because everything works just fine when using NMS to send packets. I have no idea what to do.
* Sending packets using ViaVersion API results in them not being present in the pipeline, which means plugins cannot listen to it and modify this. Some resource pack plugins do this and now they broke. Your original implementation did not even include anti-override, which I had to add later.
* As time goes by, more and more servers will update to newer MC versions, decreasing amount of servers that would benefit from any of this every day.

So pretty much since day 1 of this being released there have been various issues with it, some of which I wasn't able to solve and just more keep appearing. I belive the cost outweighs the benefit at this point. It was an interesting idea, but it failed.

Rubenicos:
I understand, seeing the situation this will only bring future problems, I implemented ViaVersion based on its behaviour on packet sending, but it seems that this differs from the game's own protocol in certain aspects, so a complete investigation of Minecraft client code would be required to make something without errors.

By removing this feature you could leave a getter and setter for `serverImplementationProvider` field in `BukkitPlatform`?
I'd at least like to make ViaVersion addon as a personal project.

NEZNAMY:
That would not solve your request, because there are things you must take from the server itself, such as player skin. But I guess you will find out. Either way, I can do that one exact thing for you if you want. Let me know when you decide all these kinds of little changes you need.

PR #1456, opened 2025-03-19
Added %luckperms-weight% as a built-in placeholder

Radu-Voinea:


NEZNAMY:
Did you really need to reformat those files?

NEZNAMY:
The changes are still there. Also you can get LP user from the tabplayer object like group does it for better performance.

Radu-Voinea:
Opened another PR #1457 as it was easier to get rid of that commit that way and not mess the git history

PR #1458, opened 2025-03-22
update rgb format #< for <#

MrRafaaeL:


NEZNAMY:
Why? This might conflict with minimessage.

PR #1464, opened 2025-03-29
Fix ViaVersion packets delay

Rubenicos:
New and improved packet sending:
* Player enter to server
* ViaVersion decoder is registered while player is on LOGIN state
* TAB plugin queue packets
* Player go to CONFIGURATION state
* Player go to PLAY state
* TAB send queued packets and start working normally


NEZNAMY:
I was hoping for a better solution than periodic checking, but I guess there's nothing better available.

Rubenicos:
The other option was making a packet listener for [LOGIN_SUCESS](https://minecraft.wiki/w/Java_Edition_protocol#Login_Success) and [FINISH_CONFIGURATION](https://minecraft.wiki/w/Java_Edition_protocol#Finish_Configuration), but inject another decoder in the pipeline just for first-time join I think is a bit excesive.

PR #1482, opened 2025-04-20
Add basic implementation of the new tab list sorting feature (1.21.2)

PapiCapi:
Hello ! 
I create this PR in order to support the new feature allowing in newer versions to sort players without teams ([Velocity](https://jd.papermc.io/velocity/3.4.0/com/velocitypowered/api/proxy/player/TabListEntry.html#setListOrder(int)), [Paper](https://jd.papermc.io/paper/1.21.5/org/bukkit/entity/Player.html#setPlayerListOrder(int)) ).

This is only a basic implementation and many things could/should be changed. Here is a small list of the choices I made or things that should be looked at : 

- [ ] The config option is currently `new-tablist-sorting`. Might be a good idea to rename it or move it to a proper config section.
- [ ] Add a "weight" system to the SortingTypes so that if SortingType A with a position at 5 and SortingType B with a position at 10 is not the same as SortingType A with a position at 10 and SortingType B with a position at 5
- [ ] Properly test each `SortingType#getPosition()`. I only tested the one I used : PlaceholderHighToLow and PlaceholderLowToHigh
- [ ] "new tab list sorting" is kind of ugly but I couldn't find a better name for each variables or config option
- [ ] Search if this option could be used on BungeeCord, Fabric, Forge, Neoforge and Sponge and implement the method `public void setTabPosition(int position)` in their TabPlayer implementations.

I really think that this new feature allowing to sort the tab list without using teams is great and should be used even if it is not the way that I am suggesting.
I allowed edits from maintainers so that you could change it the way you would've liked it :)

### DIFF
--- a/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/BukkitTabPlayer.java
+++ b/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/BukkitTabPlayer.java
@@ -2,15 +2,20 @@ package me.neznamy.tab.platforms.bukkit;
 
 import lombok.Getter;
 import lombok.SneakyThrows;
+import me.neznamy.chat.component.SimpleTextComponent;
 import me.neznamy.chat.component.TabComponent;
 import me.neznamy.tab.platforms.bukkit.hook.LibsDisguisesHook;
 import me.neznamy.tab.platforms.bukkit.platform.BukkitPlatform;
+import me.neznamy.tab.shared.TAB;
 import me.neznamy.tab.shared.backend.BackendTabPlayer;
 import org.bukkit.entity.Player;
 import org.bukkit.metadata.MetadataValue;
 import org.bukkit.potion.PotionEffectType;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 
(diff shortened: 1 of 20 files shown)


NEZNAMY:
* I haven't implemented this, because it is in fact more limiting than the teams. Even versions with 16 char limit for team names (<1.18) can fit 16*16 bits, while an integer can only hold 32 bits of data. This means combining multiple sorting types will be a major issue and may result in data loss if we want independent sorting for each player. If we go for dependent sorting (sorting players internally and them giving them numbers 1, 2 etc..), that would have a high CPU usage, as it would have to recalculate and resend all data every time someone joins, leaves, or anything regarding their current sorting value changes.
* Your setPlayerListOrder implementation is absolutely horrible - throwing errors on bukkit, not implemented in other platforms. TAB already has TabList interface, which contains an implemented setListOrder method.
* Layout uses "team names" to do exactly what I described in the point above (because there is no other way), which you now cut off. Layout will now probably not sort players correctly.

With how complicated the needs of modern gaming servers have become and how limiting this new feature is, I don't see this being implemented at any point in the future. Teams are also not bad, they work, no problems are currently present and I believe TAB is the best plugin for managing teams, so it's not like anyone would want to use a different team plugin while wanting TAB to sort players (I hope?).

NEZNAMY:
Maybe one day I might decide to implement this in some way, but there currently no plans.

Issue #1465, opened 2025-03-30
TAB List Sort Issue


CowsAreCool26:
### Server version

Velocity 3.4.0 Latest

### TAB version

5.1.0

### Plugin list

velocity, commandsync, connectorplugin, easycommandblocker, fastmotd, floodgate, luckperms, geyser, forceresourcepacks, geyserutils, networkjoinmessages, openaudiomc, redisbungee, parties, premiumvanish, signedvelocity, skinsrestorer, velocity-scoreboard-api, tab, vulcan

### Describe the bug

I'm running multiproxy support through Redis, users randomly do not sort properly. For example 

![Image](https://github.com/user-attachments/assets/e90251c4-2bf1-41d2-88d4-beadff4ae962)

Here's the debug about the user
 btab debug .Malachi173623
[15:50:16 INFO] [TAB]: [TAB] Showing debug information
[15:50:16 INFO] [TAB]: >-------------------------------<
[15:50:16 INFO] [TAB]: Server version: [Velocity] Velocity - 3.4.0-SNAPSHOT (git-d2cd7918-b483)
[15:50:16 INFO] [TAB]: Plugin version: 5.1.0
[15:50:16 INFO] [TAB]: Permission plugin: LuckPerms
[15:50:16 INFO] [TAB]: Permission group choice logic: Primary group
[15:50:16 INFO] [TAB]: Sorting system: GROUPS -> PLACEHOLDER_A_TO_Z
[15:50:16 INFO] [TAB]: Storage type: File
[15:50:16 INFO] [TAB]: >-------------------------------<
[15:50:16 INFO] [TAB]: Player: .Malachi173623
[15:50:16 INFO] [TAB]: Player version: 1.21.4
[15:50:16 INFO] [TAB]: Bridge connection: Connected
[15:50:16 INFO] [TAB]: Primary permission group: newbie
[15:50:16 INFO] [TAB]: Team name: i.malachi173623A
[15:50:16 INFO] [TAB]: Sorting note:
-> Primary group (newbie) is #58 in sorting list.
-> %player% returned ".Malachi173623".
[15:50:16 INFO] [TAB]: tabprefix: "%luckperms-prefix% " (Source: group=newbie)
[15:50:16 INFO] [TAB]: customtabname: "%player%" (Source: group=_DEFAULT_)
[15:50:16 INFO] [TAB]: tabsuffix: "%essentials_afk%" (Source: group=_DEFAULT_)
[...]

PR #1485, opened 2025-04-28
Full messenger implementation

Rubenicos:
> [!WARNING]
> This pull request is not yet tested.

Related to #1465

Instead of using a built-in messenger executor from delivery4j (that is affected by multi-threading enviroments), this pull request use a properly made ThreadExecutor and also use TAB console logger.


NEZNAMY:
Already solved in a different way.

PR #1554, opened 2025-10-01
Add customnametag support for placeholder-based player names in nametags [example]

GG-MD:
### Summary
Adds `customnametag` to allow placeholder-based names in nametags, similar to `customtabname` for tablist.

### Problem
TAB used `player.getNickname()` for nametags, so placeholders like `%cmi_user_display_name%` didn’t work. Colors showed in tablist but not above players.

### Solution
- Added `customnametag` to `NameTagPlayerData.java`
- Updated `NameTag.java` to load/use `customnametag` instead of `player.getNickname()`
- Updated `NameTagProxyPlayerData.java` for proxy support
- Added `customnametag: "%cmi_user_display_name%"` to `groups.yml`

### Changes
1. **NameTagPlayerData.java**: Added `customName` property
2. **NameTag.java**:
   - `loadProperties()` loads `customnametag`
   - `updateProperties()` updates `customnametag`
   - `registerTeam()` uses `customName` when available
   - `refresh()` handles `customName` updates
3. **NameTagProxyPlayerData.java**: Added `customName` to proxy messages
4. **groups.yml**: Added `customnametag: "%cmi_user_display_name%"`

### Benefits
- Placeholders work in nametags
- Colors from `%cmi_user_display_name%` display above players
- Backward compatible
- Consistent with `customtabname`

### Testing
- Colors display in tablist and nametags
- `%cmi_user_glow_code%` works with custom names
- Proxy support works
- No regressions

### Usage
```yaml
_DEFAULT_:
  customnametag: "%cmi_user_display_name%"
```

This enables placeholder-based names with colors in nametags, resolving the issue where colors appeared in tablist but not above players.

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java
@@ -105,6 +105,9 @@ public class NameTag extends RefreshableFeature implements NameTagManager, JoinL
             refresh = prefix || suffix;
         }
         if (refresh) updatePrefixSuffix(refreshed);
+        if (refreshed.teamData.customName != null && refreshed.teamData.customName.update()) {
+            updatePrefixSuffix(refreshed);
+        }
     }
 
     @Override
@@ -217,6 +220,7 @@ public class NameTag extends RefreshableFeature implements NameTagManager, JoinL
     private void loadProperties(@NotNull TabPlayer player) {
         player.teamData.prefix = player.loadPropertyFromConfig(this, "tagprefix", "");
         player.teamData.suffix = player.loadPropertyFromConfig(this, "tagsuffix", "");
+        player.teamData.customName = player.loadPropertyFromConfig(this, "customnametag", null);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Tanguygab:
where is the player name actually changed? the only thing you did is load the customtagname property and use it in the scoreboard team packet instead of the player's nickname, which won't work since scoreboard teams only dictate prefix and suffix, not name.

TAB doesn't modify players' usernames, though it does listen to other plugins tua5 do, which is why the scoreboard team is sent with the player's nickname instead of his real username.

Tl;dr TAB doesn't change player names, your PR doesn't either and we have no plans to do so. Use a nickname plugin that does and TAB will listen to name changes.

GG-MD:
> where is the player name actually changed? the only thing you did is load the customtagname property and use it in the scoreboard team packet instead of the player's nickname, which won't work since scoreboard teams only dictate prefix and suffix, not name.
> 
> TAB doesn't modify players' usernames, though it does listen to other plugins tua5 do, which is why the scoreboard team is sent with the player's nickname instead of his real username.
> 
> Tl;dr TAB doesn't change player names, your PR doesn't either and we have no plans to do so. Use a nickname plugin that does and TAB will listen to name changes. 

The problem is that cmi_user_displayname doesn’t show in HEX/GRADIENT color format above the player’s head — it only shows correctly in the tab list. If the issue were with CMI itself, then colored nicknames wouldn’t appear in the tab list either, but that’s not the case.

Tanguygab:
Nametags don't support RGB colors as stated on the wiki. This is neither a CMI or a TAB issue. 
And this still doesn't change the fact that your PR doesn't edit the player's username with the colored CMI nickname, so the result is the same.

PR #1595, opened 2025-12-27
feat: add RedisBungee placeholder support for network player count

gbpii:
- Add "%redisbungee_total%" placeholder to BungeeCord and Velocity platforms
- Implement automatic RedisBungee detection and placeholder registration
- Configure default 1000ms refresh interval for network player count
- Include error handling and debug logging for RedisBungee integration
- Update shared config template with placeholder refresh intervals

This enables TAB users to display total network player count across all proxies in their tablist, header/footer, and scoreboard configurations when using RedisBungee-compatible plugins.


Tanguygab:
or just use %online% while proxy-support is enabled?

gbpii:
I was testing both, with our setup sometimes online breaks up, shows more (like +1) or waits little more than placeholder that i added.

JVM temurin 25
Multiple velocities & redis

NEZNAMY:
For delay, decrease refresh interval.
If the count is objectively wrong, it suggests there is a player in memory who should not be there, which is a bigger issue that needs proper fixing rather than closing your eyes before it.

gbpii:
Thanks! Thats a good one, lets close this up then.
(Will check every necessary place)

PR #1598, opened 2026-01-07
Layout: allow placeholder-based ping values for fixed slots

Mattrolol:
## Summary
- Allow `fixed-slots` to accept placeholder ping values (4th segment) instead of numeric-only.
- Resolve ping values per viewer and update latency bars dynamically.

## Details
Previously, the 4th segment (`PING`) in `fixed-slots` only accepted numeric literals. This change stores the raw ping string, resolves placeholders via `Property`, and updates tab latency when the placeholder changes. Invalid/non-numeric values fall back to `layout.empty-slot-ping-value`.

## Example
Before (worked only with numeric):
`1|Text|skin|1000`

Now (placeholder supported):
`1|Text|skin|%server_ping%`

## Compatibility
- Backward compatible for numeric pings.
- API fixed slots unchanged (still numeric ping).

## Rationale
Enables fixed-slot ping bars to be driven by live placeholders (server status/health/proxy metrics) instead of hardcoded numbers, while keeping numeric pings supported.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -11,7 +11,7 @@ dependencyResolutionManagement {
         maven("https://repo.opencollab.dev/maven-snapshots/") // Floodgate, Bungeecord-proxy
         maven("https://repo.purpurmc.org/snapshots") // Purpur
         maven("https://jitpack.io") // PremiumVanish, Vault, YamlAssist, RedisBungee
-        maven("https://mvn.lib.co.nz/public/") // LibsDisguises
+        maven("https://repo.md-5.net/content/groups/public/") // LibsDisguises
         maven("https://repo.william278.net/velocity/") // Velocity-proxy
     }
 }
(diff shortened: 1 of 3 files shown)


settings.gradle.kts
```diff
@@ -11,7 +11,7 @@ dependencyResolutionManagement {
         maven("https://repo.opencollab.dev/maven-snapshots/") // Floodgate, Bungeecord-proxy
         maven("https://repo.purpurmc.org/snapshots") // Purpur
         maven("https://jitpack.io") // PremiumVanish, Vault, YamlAssist, RedisBungee
-        maven("https://repo.md-5.net/content/groups/public/") // LibsDisguises
+        maven("https://mvn.lib.co.nz/public/") // LibsDisguises
```
NEZNAMY: Why this?

NEZNAMY:
Really good PR, thanks! Just to confirm, you have tested it all, correct?

Mattrolol:
Yes, I tested it on my end, and it seems to be working as expected.

NEZNAMY:
Can you address the comments and then test the new changes?

Mattrolol:
Ok, I addressed the notes, tested the changes, and everything works as expected.

NEZNAMY:
Thanks again!

PR #1599, opened 2026-01-10
feat: better wildcard support and enhanced pettern matching w/ regex for global player list

byPixelTV:
### Motivation
- I'm using a cloud system to dynamically start and stop new servers. The problem is that if you don't have a static amount of lobbys you had to put way to mutch entries of servers in the config and even if you would do that, there is still the possibility that you create more lobbys than specified in the config. 

### Modification
- Updated repo for LibsDisguises in `settings.gradle.kts` because Gradle could not resolve the dependency, it was not in any repo specified in the project before the change. If you try to access `https://repo.md-5.net/content/groups/public/` you get a 404, so it had to be changed so Gradle can correctly fetch the dependency.
- Added the `matchesServerPattern()` method in `DataManager` to centralize the pattern matching process. Now wildcards are case-sensitive and other pattern matching methods are added
- Updated player counting logic in `GlobalPlayerList` to also use the new pattern matching system

### Result
- Wildcards are now case-sensitive
- You can use the new pattern matching methods for more fine control which servers should share their tablist
- The wiki is updated to properly document this feature

### DIFF
--- a/.github/wiki/feature_guide_global_playerlist.md
+++ b/.github/wiki/feature_guide_global_playerlist.md
@@ -67,6 +67,36 @@ To make all servers share the playerlist, simply clear server groups and set
   server-groups: {}
 ```
 
+### Pattern Matching
+Server names in groups support multiple pattern types for flexible configuration:
+
+**Exact Match** - Matches server name exactly:
+```yaml
+server-groups:
+  lobbies:
+    - lobby1
+    - lobby2
+```
+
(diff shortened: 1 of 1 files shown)


NEZNAMY:
Thanks for the contribution! 

Though I believe things already were case sensitive, so this doesn't change anything, unless I am missing something.

About the wiki part: The wiki in source code is a backup manually taken once per month. I will update the actual wiki with your new content when this comes out (if a new TAB version does not come out by the next backup, I will make sure to not overwrite your changes).

PR #1600, opened 2026-01-11
perf: replace String.format with StringBuilder for faster placeholder…

groundbreakingmc:
Replaces `String.format()` with `StringBuilder` for manual placeholder substitution in `update()`.
Screenshot at Jan 11 04-42-49

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/Property.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/Property.java
@@ -20,6 +20,9 @@ import java.util.List;
  */
 public class Property {
 
+    /** Thread-local StringBuilder pool for efficient string building without allocations */
+    private static final ThreadLocal<StringBuilder> STRING_BUILDER_POOL = ThreadLocal.withInitial(() -> new StringBuilder(256));
+
     /** Internal identifier for this text for PlaceholderAPI expansion, null if it should not be exposed */
     @Getter
     @Nullable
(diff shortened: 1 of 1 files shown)


NEZNAMY:
Can you provide old and new spark reports for the same time period? Both CPU and memory allocations.

While you're at it, you can also replace
```
if ("%s".equals(rawFormattedValue)) {
```
with a new variable such as `singlePlaceholderOnly` assigned in `analyze()`.

NEZNAMY:
After thinking about it further, we can skip this almost entirely. We can replace rawFormattedValue with an array, which would consist of 2 options: literal string and placeholder. Then, each element would be either appended literally, or returned as placeholder. No need for `%s` or any sort of string index stuff.

Would you like to do this or will you leave it to me?

NEZNAMY:
Did you use AI for this? It was way too fast, I didn't even get to finish my version yet. 

Also I am not really sure about the string builder pool. I would save initialCapacity as a field in analyze(), with value being static text size + 16*placeholder count.

I believe the "placeholders" variable is no longer used. 

Regarding your "Single literal element - fast path", it should return false immediately. That is what the current code does.

groundbreakingmc:
I did this a few hours ago. I figured out how to split things into chunks almost after the first commits.

And only addded Elements before make last commit.

But I wrote the docs and comments with AI (I don't speak English well enough).

groundbreakingmc:
and you can edit my commits, so remove unused and imvprove what you want

groundbreakingmc:
https://spark.lucko.me/G6AVk8guVv?hl=6248
https://spark.lucko.me/h4FLcK9bpo?hl=21792

NEZNAMY:
Thanks for the idea and initial implementation, I will fine-tune it myself now.

PR #1616, opened 2026-02-11
fix/not-working-translations-in-playerlist-prefixes-and-suffixes

EarWarm:
The fix is based on rendering the translation components before setting the player's display name in the tab list entry. 

By default, Velocity performs rendering when sending for the header and footer tablist, but it does not do this for the player entry in the player list.

My fix may not be perfect for the codebase of this plugin, my task was to show a working solution to the problem.

My fix is made only for Velocity, but it can also be used for Paper, I don't have enough immersion in the code base of the TAB plugin, so I hope that the fix will be expanded to a full-fledged fix on all platforms where it will be required.

Before fix:
https://drive.google.com/file/d/1u6Yvh2rUE0eZMaitir_Gev9ugUe4p8Ln/view?usp=sharing

After fix:
https://drive.google.com/file/d/1s48g-Hllz8vVGpxn3UX2amUbRibTYwJU/view?usp=sharing


NEZNAMY:
* This runs even if there are no translatable components, drastically increasing CPU usage.
* This still does not support other features - nametags, belowname number format, playerlist objective number format, sidebar scoreboard, bossbar.

PR #1626, opened 2026-03-01
[Placeholders] Added locale of date-/time-format to config.yml

BlvckBytes:
Greetings!

I was looking to use a `date-format` of `E dd.MM.yyyy` on our server when I noticed that the English locale was hard-coded - this little patch adds a config-option such that, for example, the name of the current day (`E`) can be adapted as needed.

I've tried to adhere to the existing style and hope to not have missed anything crucial - if so, please let me know and I'll get right to it. Thanks!


NEZNAMY:
Thanks for the contribution. I will add config conversion that adds this option into existing configs.

NEZNAMY:
I have tested this and the warning was only printed if the configured locale is an empty string. If I set it to something invalid, it's not printed. The check needs to be changed.

Issue #1628, opened 2026-03-09
Scoreboard Score sent about Vanished players


c0bl:
### Server version

Paper 1.21.11

### TAB version

v5.5.0 and v6.0.0-SNAPSHOT

### Plugin list

[01:17:01 INFO]:  - ChatManager, FreedomChat, GrimAC, LuckPerms, packetevents, mcCore, mcdc, SayanVanish, TAB, WorldEdit

### Describe the bug

While Vanished players are properly hidden from the TAB scoreboard list it was noticed that normal players without any permission to see other vanished players are sent `ClientboundSetScorePacket` packets containing the name and score for vanished players.

These packets are sent from the YellowNumber feature and can be used by malicious players to defeat a vanish plugin.

### Expected behavior

TAB should use the `canSee` method before sending player information out.

### Steps to reproduce

Step 1. Install TAB
Step 2. Configure the `playerlist-objective` to be enabled with any value.
Step 3. Install SayanVanish or comparable vanish plugin
Step 4. Vanish in game on admin account
Step 5. Notice that normal user receives ClientboundSetScorePacket packets with admin username

### Checklist

- [x] I am running latest version of the plugin
- [x] I have read the wiki to make sure it's not an issue with configuration
- [ ] I ticked all of the boxes without actually reading them
- [x] I have read the Compatibility wiki page and am not trying to run the plugin on an unsupported server version / platform
- [x] I am willing to cooperate to get this issue resolved faster, including providing any additional information requested or testing new builds

PR #1629, opened 2026-03-09
Consider canSee in YellowNumber.java setScore

c0bl:
Fixes https://github.com/NEZNAMY/TAB/issues/1628

consider `canSee` when sending scoreboard score packets in `YellowNumber` feature


NEZNAMY:
Thanks for the contribution. I will finish this to properly update the value once player unvanishes, as well as add this check to its sibling feature - Belowname.

PR #1650, opened 2026-05-01
[Belowname] Implement disable-condition as target since 26.2-snapshot-5 now allows it

NEZNAMY:
In 26.2-snapshot-5, Mojang fixed [MC-307012](https://bugs.mojang.com/browse/MC/issues/MC-307012) and now default NumberFormat does not appear on anyone anymore. Entities (including players) without a score no longer have any belowname. 

This allows a new feature - disabling belowname for target players. 

This PR implements that. It is unclear whether this is intended or not and will be fixed, so for now, keeping it as a PR until 26.2 is released.

This PR also contains a TODO to only send reset packet to 26.2+ players to prevent players below this version from seeing default value on players, which is never desired.


NEZNAMY:
Turns out there is a much easier way to implement this.

Issue #1648, opened 2026-04-25
Compatibility concern with future versions of Java


NEZNAMY:
### TAB version

6.0.1

### Software causing compatibility issue

In Java 26, it was [announced](https://docs.oracle.com/en/java/javase/26/migrate/preparing-final-field-mutation-restrictions.html) that in future releases, it will no longer be possible to modify `final` fields using reflection. This is something TAB uses in multiple places.
Running TAB on Java 26 produces the following warning in the console:
```
> WARNING: Final field players in class net.minecraft.network.protocol.game.ClientboundSetPlayerTeamPacket has been mutated reflectively by class me.neznamy.tab.platforms.bukkit.paper_1_21_11.NMSPacketScoreboard in unnamed module @5e875754 (file:<path>)
WARNING: Use --enable-final-field-mutation=ALL-UNNAMED to avoid a warning
WARNING: Mutating final fields will be blocked in a future release unless final field mutation is enabled
```
With possibly more classes and fields when triggered.
This warning can be disabled by adding the following option to startup arguments:
```
--illegal-final-field-mutation=allow
```
Alternatively, adding this:
```
--illegal-final-field-mutation=deny
```
Will cause every attempt to throw an error.

This issue is affecting a lot of big plugins as fields being private and final is very frequent in the Minecraft code. It may not be possible to resolve some of them.

According to the announcement, explicitly allowing final field mutation will have to be enabled using a command line argument. This is most likely going to be the solution the Minecraft community will adopt. The purpose of this issue is to bring this up and see if something can be reasonably done about it.

### Additional info

TAB uses reflection to change final fields in the following places:
* `entries` field in the player info packet - creating the packet due to a lack of necessary constructor (this was added by Paper), modifying the list to replace entries as modifications are not possible due to the player data class being a record.
* `players` field in the team packet - remove players from packets coming from other plugins so they don't break TAB

Proxies are unaffected - all packets have setters and their fields are not final.

Steps towards resolving this:
- [x] Use access transformers on modded platforms to remove `final` modifier from all fields (and also make them `public` at that point)
- [ ] Find ways to update team packet players collection:
  - [ ] 1.10 - 1.16.5 - The `players` field is private and final, but the collection is mutable (`Lists.newArrayList()`). Update the collection. 
  - [ ] 1.17+ - Has an all-arg contructor (which is private), with the `players` field being immutable. Replace the entire packet.
 - [ ] Find ways to update player collection in player info packet:
[...]

PR #1659, opened 2026-05-26
Fix Canvas async scoreboard team packets

DimaSergeew:
## Summary

Fixes scoreboard team packet creation on Canvas/Paper 1.21.11 by avoiding mutation of `PlayerTeam` state when TAB only needs to send a packet snapshot.

Canvas enforces scoreboard team property changes on the global tick thread. TAB refresh tasks can run asynchronously, so calling `PlayerTeam` setters such as friendly fire, visibility, collision, prefix, suffix, or color can throw `IllegalStateException`.

This changes the Paper 1.21.11 scoreboard implementation to create `ClientboundSetPlayerTeamPacket` snapshots directly and populate packet parameters without mutating scoreboard team state.

## Validation

- Built `:bukkit:paper_1_21_11:build`
- Built `:bukkit:paper_1_21_11:shadowJar`
- Confirmed the patched jar starts on Canvas without the async friendly-fire/team-property crash


NEZNAMY:
This idea goes directly against #1648 . 
I was about to solve it by scheduling these tasks in the main tick thread on Canvas.

Issue #1662, opened 2026-05-27
CanvasPacketScoreboard crashes on disable - tries to schedule task after plugin disabled


Potenza7:
## what's happening

when the server shuts down on canvas (folia fork), TAB crashes during `onDisable` because `CanvasPacketScoreboard.unregisterTeam()` tries to dispatch to the global region scheduler but the plugin is already marked as disabled at that point. folia straight up rejects it with `IllegalPluginAccessException: Plugin attempted to register task while disabled`.

basically the shutdown flow goes like this:
1. `RegionShutdownThread` starts disabling plugins
2. TAB's `onDisable` fires → `TAB.unload()` → `SafeScoreboard.clear()` → `unregisterTeam()`
3. `CanvasPacketScoreboard.unregisterTeam()` calls `FoliaPlatform.runSyncGlobal()` to dispatch the packet on the global tick thread
4. `FoliaGlobalRegionScheduler.execute()` says no and  plugin is disabled, throws `IllegalPluginAccessException`

## version info

- TAB v6.1.0-SNAPSHOT (latest dev build)
- Canvas (26.1.2)
- Java 25

## full stack trace

```
[19:23:32 INFO]: [TAB] Disabling TAB v6.1.0-SNAPSHOT
[19:23:32 INFO]: [TAB] [WARN] Failed to disable
[19:23:32 INFO]: [TAB] [WARN] java.lang.reflect.InvocationTargetException: null
[19:23:32 INFO]: [TAB] [WARN]   at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:119)
[19:23:32 INFO]: [TAB] [WARN]   at java.base/java.lang.reflect.Method.invoke(Method.java:565)
[19:23:32 INFO]: [TAB] [WARN]   at me.neznamy.tab.platforms.bukkit.platform.FoliaPlatform.runSyncGlobal(FoliaPlatform.java:170)
[19:23:32 INFO]: [TAB] [WARN]   at me.neznamy.tab.platforms.bukkit.paper_1_21_11.CanvasPacketScoreboard.unregisterTeam(CanvasPacketScoreboard.java:56)
[19:23:32 INFO]: [TAB] [WARN]   at me.neznamy.tab.shared.platform.decorators.SafeScoreboard.unregisterTeam(SafeScoreboard.java:158)
[...]

PR #1664, opened 2026-05-30
Improve scheduler handling

R00tB33rMan:
The `GlobalScheduler` here shouldn't need a (plugin is enabled) check, and `runSync` under an `EntityScheduler` should use `execute`, which inherently checks if the plugin is enabled and fully completes the operation.

Let me know if any other changes are needed!


NEZNAMY:
#1662 disagrees. What changes were needed in the first place? Everything is working.

R00tB33rMan:
Going to void this and if I run into issues, I'll let you know!

Issue #1233, opened 2024-02-17
Feature request: MiniPlaceholders support


Kamillaova:
### Describe the feature request

MiniPlaceholders is a _**MiniMessage Component-based Placeholders for Paper, Fabric, Sponge, Folia and Velocity platforms**_

It would be great to see MiniPlaceholders support in TAB

### Additional info

https://github.com/MiniPlaceholders/MiniPlaceholders
https://github.com/MiniPlaceholders/MiniPlaceholders/wiki/Projects-using-MiniPlaceholders-API

NEZNAMY:
[Feature requests are closed](https://github.com/NEZNAMY/TAB?tab=readme-ov-file#provided-services). What's the point of doing this anyway? PlaceholderAPI is available on bukkit and with bridge you can display placeholders on bungee & velocity as well. That covers roughly 99.98% of TAB users. The placeholders are also basic and don't seem to offers hooks into plugins.

Kamillaova:
> What's the point of doing this anyway?

\*1.16+\* placeholders with a fancy implementation, not through smth like `UNUSUAL_LEGACY_COMPONENT_SERIALIZER.serialize(component)`

That's all.

NEZNAMY:
I am not struggling with serializing components on 1.16+.

Kamillaova:
> I am not struggling with serializing components on 1.16+.

This is about placeholders developers, also about PlaceholderAPI in general

NEZNAMY:
I see no reason to do this even if feature requests were open, which they are not. PlaceholderAPI offers everything you need and much more. Supporting more than 1 placeholder source can also be problematic.

Kamillaova:
Okay

PR #1677, opened 2026-06-24
Add MiniPlaceholders support for Velocity

Crafting-tunbol:
Adds optional MiniPlaceholders support to the Velocity module. The integration is a soft dependency and only activates when MiniPlaceholders is present. No changes to Bukkit, BungeeCord, PlaceholderAPI, or TAB-Bridge behavior.

Issue #1233 highlights a gap for Velocity-only networks. Proxy-only setups currently have no placeholder source unless TAB-Bridge is deployed across backend servers, adding unnecessary deployment and maintenance requirements for networks that do not otherwise need backend plugins. MiniPlaceholders provides a native Velocity-side placeholder ecosystem, including integrations such as LuckPerms, Vault, and ProxyConnection.

MiniPlaceholders 3.x requires Java 21. To support the integration, the Velocity module target is updated from Java 17 to Java 21.
This change only affects the Velocity module and aligns with the current Velocity platform direction, as newer Velocity releases already require Java 21.


NEZNAMY:
Looking really good. 

To my understanding, this makes miniplaceholders take priority over bridge for both providing and parsing placeholders, correct? What if someone installs miniplaceholders for other purposes and still wants backend PlaceholderAPI support? This case seems to not be handled.

So this internally changes all `%placeholder%` requests to `<placeholder>` and passes it to miniplaceholders?

velocity/src/main/java/me/neznamy/tab/platforms/velocity/hook/MiniPlaceholdersHook.java
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Converts an Adventure component into legacy text supported by TAB.
+     *
+     * @param   component
+     *          component to convert
+     * @return  legacy text representation of the component
+     */
+    @NotNull
+    public static String toLegacyString(@NotNull Component component) {
+        return LEGACY_SERIALIZER.serialize(component);
```
NEZNAMY: Perhaps you can take inspiration from https://github.com/NEZNAMY/TAB/blob/master/fabric/src/main/java/me/neznamy/tab/platforms/fabric/hook/PlaceholderAPIHook.java#L38-L65 to preserve RGB colors.

Crafting-tunbol:
I hadn't looked closely enough at how TAB-Bridge is currently used, and you're right, my initial implementation would have conflicted with Bridge placeholders.
My original idea was to transparently convert %placeholder% to MiniPlaceholders syntax to make migration essentially seamless for Velocity-only users. However, that breaks down when someone is using both MiniPlaceholders and TAB-Bridge/PlaceholderAPI at the same time.

I've changed it so MiniPlaceholders only processes its native <placeholder> syntax. %placeholder% remains untouched and continues to be handled by TAB-Bridge/PAPI as before. This allows both systems to coexist without conflicts and gives users explicit control over which provider they want to use. I also adjusted the legacy color handling, which should behave better now.

One thing worth noting is that some placeholders can still produce different output between PAPI and MiniPlaceholders because the formatting syntaxes differ. For example, a value containing:
`<#E0B11E>MyServer</#FF0000>`
won't render correctly when returned through MiniPlaceholders, since MiniPlaceholders expects:
`<gradient:#E0B11E:#FF0000>MyServer</gradient>`

Example in a scoreboard:
```yaml
- "<#E0B11E>MyServer</#FF0000>"
- "PAPI Bridge      : %luckperms_meta_normalgradient%"
- "MiniPlaceholders : <luckperms_meta:normalgradient>"
- ""
- "<gradient:#E0B11E:#FF0000>MyServer</gradient>"
- "PAPI Bridge      : %luckperms_meta_fullgradient%"
- "MiniPlaceholders : <luckperms_meta:fullgradient>"
```

demo-scoreboard-gradient-rendering-issue-miniplaceholders

NEZNAMY:
I don't understand why it doesn't work. The first format is TAB's and second is just plain MM, but both are processed after placeholders. Were you able to find out why this breaks?
Breaking the 2 main gradient syntaxes is quite problematic.

Crafting-tunbol:
Just to clarify, in the screenshot I intentionally replaced `#` with `＃` to escape the gradients and display both syntaxes side by side in the example. The normal TAB/MM syntax (`<#E0B11E>MyServer</#FF0000>`) still works perfectly fine, both directly and through PAPI Bridge. Nothing is broken there.
The issue only appears when MiniPlaceholders returns a value containing `<#E0B11E>MyServer</#FF0000>`. In that case it renders like in the screenshot instead of being parsed correctly. I haven't tracked down the exact cause yet, but I'll investigate further and see where the formatting is getting lost.

NEZNAMY:
Oh I see, that explains why the symbol looks different in the config. I didn't know there's another symbol that looks the same.

Good luck investigating, it looks really good so far. You even managed to solve my biggest fear - supporting both `%%` and `<>`.

Crafting-tunbol:
MiniPlaceholders returns an already parsed Adventure Component. By the time TAB receives it, a tag like `<#E0B11E>` has already been consumed and converted into component styling information, so the original MiniMessage syntax no longer exists for TAB to process.
With PAPI, the placeholder returns a raw string, so TAB still receives `<#E0B11E>...</#FF0000>` and can parse it normally.

`<gradient:#E0B11E:#FF0000>...</gradient>` syntax appears to work because MiniPlaceholders doesn't interpret that syntax itself. It passes it through as plain text, allowing TAB to parse it later.
In theory, the only workaround I can think of would be inspecting the returned Component and trying to reconstruct MiniMessage syntax from it, but that feels fragile and likely wouldn't be lossless in all cases.
I'm not sure how common it is for placeholders themselves to return gradient tags, but this is definitely a limitation of the current approach. If users run into it, using the <gradient:...> syntax would avoid the issue since it bypasses MiniPlaceholders and is still parsed by TAB afterwards.

NEZNAMY:
When `<#E0B11E>` is parsed, what does the returned component look like? Isn't it something like `{"text": "<#E0B11E>"}`?

Crafting-tunbol:
No, actually, it looks more like:
`<#E0B11E>MyServer</#FF0000>`
becomes
`TextComponentImpl{content="MyServer</#FF0000>", style=StyleImpl{obfuscated=not_set, bold=not_set, strikethrough=not_set, underlined=not_set, italic=not_set, color=TextColorImpl{value="#E0B11E"}, shadowColor=null, clickEvent=null, hoverEvent=null, insertion=null, font=null}, children=[]}`
So you are losing the tag here.

MiniPlaceHolders use miniMessage.deserialize(), which gives this result
`miniMessage().deserialize(providedString, source, resolver);`
See: https://github.com/MiniPlaceholders/MiniPlaceholders/blob/main/common/src/main/java/io/github/miniplaceholders/common/command/node/ParseNode.java#L24-L41

NEZNAMY:
So using TAB's gradient syntax in miniplaceholders is broken? I guess if people want to use miniplaceholders, they may as well use MM syntax (I bet miniplaceholers enjoyers will refuse to use anything other than MM anyway).

Crafting-tunbol:
Yes, TAB's gradient syntax is broken using MiniPlaceHolders because of their mini message pre-processing.
It's just a different design compared to PAPI, which keeps everything as raw strings until the consumer decides what to do with them. If PAPI had originally adopted this approach, I think it would have made sense as well.

A warning in the `how_to_use_minecraft_components` wiki page would be necessary for MiniPlaceholders users, with a short explanation of this behavior and a link to this PR for the technical details.

Would you prefer putting it in the `RGB colors (1.16+)` section, or in the `MiniMessage` section?

Something along these lines:

> [!WARNING]
> TAB's gradient syntax (`<#RRGGBB>Text</#RRGGBB>` and similar) relies on TAB preprocessing the input before MiniMessage parsing. On **Velocity**, this can interact with MiniPlaceholders because MiniPlaceholders returns an already parsed Adventure Component rather than the original MiniMessage string. As a result, placeholders that output TAB's gradient syntax cannot be reprocessed by TAB. Native MiniMessage gradients (`<gradient:#RRGGBB:#RRGGBB>Text</gradient>`) are unaffected. See [TAB integration PR #1677](https://github.com/NEZNAMY/TAB/pull/1677) for more details.

NEZNAMY:
I would probably extend the `placeholders` page to add miniplaceholders alongside PlaceholderAPI, with this warning. Or maybe both pages would work.

Crafting-tunbol:
Done, this should be good now.

.github/wiki/placeholders.md
```diff
@@ -143,6 +144,12 @@ Full list of placeholders (general + feature specific):
 | `%tab_customtabname_raw%` | Player's current raw customtabname with placeholder identifiers. |
 | `%tab_tabsuffix_raw%`     | Player's current raw tabsuffix with placeholder identifiers.     |
 
+# MiniPlaceholders
+On **Velocity**, TAB integrates with [MiniPlaceholders](https://github.com/MiniPlaceholders/MiniPlaceholders) when the plugin is installed. TAB automatically detects and resolves MiniPlaceholders placeholders in config text, including global, per-player, and relational ones. Placeholders use angle brackets, with the expansion and name separated by an underscore (for example `<luckperms_prefix>`), and arguments added after a colon (for example `<luckperms_meta:key>`). See the [MiniPlaceholders placeholder list](https://miniplaceholders.github.io/docs/user-guide/Placeholders) for all available placeholders. TAB also registers its own expansion, offering placeholders such as `<tab_tabprefix>` and `<tab_placeholder:name>`. The `%placeholder%` syntax remains reserved for [PlaceholderAPI](https://github.com/PlaceholderAPI/PlaceholderAPI) (via PAPIProxyBridge on Velocity).
```
NEZNAMY: PAPIProxyBridge is a plugin that has nothing to do with TAB.

Crafting-tunbol: yeah my bad I meant for TAB Bridge

.github/wiki/placeholders.md
```diff
@@ -145,10 +145,10 @@ Full list of placeholders (general + feature specific):
...
+> TAB's gradient syntax (`<#RRGGBB>Text</#RRGGBB>`) is a compatibility layer that TAB rewrites before parsing, it is more fragile than native [MiniMessage](https://docs.advntr.dev/minimessage/format.html) gradients (`<gradient:#RRGGBB:#RRGGBB>Text</gradient>`). On **Velocity**, this is especially problematic because [MiniPlaceholders](https://github.com/MiniPlaceholders/MiniPlaceholders) parses MiniMessage *before* TAB can rewrite its gradient syntax. For example, placeholder output like `<#E0B11E>MyServer</#FF0000>` is interpreted by MiniPlaceholders as a hex color tag (`<#E0B11E>`), but `</#FF0000>` is not recognized as a gradient closing tag, so the formatting breaks. Prefer native MiniMessage syntax in placeholder output (e.g. `<gradient:#E0B11E:#FF0000>MyServer</gradient>`), or set `minimessage-support: false` in **config.yml** if you only need TAB's own RGB/gradient handling. See [How to use Minecraft components - MiniMessage](https://github.com/NEZNAMY/TAB/wiki/How-to-use-Minecraft-components#minimessage) for more information.
```
NEZNAMY: Will disabling MM support actually fix this?

Crafting-tunbol: I initially thought disabling MM support would help, but after testing it doesn't. MiniPlaceHolder still parses it anyway, so there's no real way to bypass it. My mistake.

NEZNAMY: Your MM hook is also not bound to the config variable, I don't know if that was the cause or you disabled it in the code to test.

Crafting-tunbol: I disabled it in the code to test, but it wouldn't make sense anyway, due to MiniPlaceHolder already parsing it.

shared/src/main/java/me/neznamy/tab/shared/features/PlaceholderManagerImpl.java
```diff
@@ -257,14 +258,29 @@ public void load() {
...
      *
      * @param   text
      *          text to detect placeholders in
      * @return  list of detected identifiers
      */
     @NotNull
     public static List<String> detectPlaceholders(@NonNull String text) {
+        List<String> placeholders = new ArrayList<>(detectPercentPlaceholders(text));
+        TAB tab = TAB.getInstance();
+        if (tab != null) {
```
NEZNAMY: This is never null.

shared/src/main/java/me/neznamy/tab/shared/placeholders/types/TabPlaceholder.java
```diff
@@ -59,11 +60,18 @@ public abstract class TabPlaceholder implements Placeholder {
...
+        if (!PlaceholderIdentifier.isValid(identifier))
+            throw new IllegalArgumentException("Identifier must start and end with % or <> (attempted to use \"" + identifier + "\")");
         this.identifier = identifier;
         this.refresh = refresh;
         Map<Object, Object> map = TAB.getInstance().getConfiguration().getConfig().getReplacements().getValues().get(identifier);
+        if (map == null) {
+            map = TAB.getInstance().getConfiguration().getConfig().getReplacements().getValues().get(
+                    PlaceholderIdentifier.toPercentSyntax(identifier).equals(identifier)
+                            ? PlaceholderIdentifier.toAngleBracketSyntax(identifier)
+                            : PlaceholderIdentifier.toPercentSyntax(identifier)
```
NEZNAMY: Is this all needed? Can't we just allow / force users to use `<>` for MP placeholders in replacements?

shared/src/main/java/me/neznamy/tab/shared/placeholders/PlaceholderRefreshConfiguration.java
```diff
@@ -33,10 +34,15 @@ public class PlaceholderRefreshConfiguration {
...
-        if (identifier.startsWith("%permission:")) { // Bad code, maybe fix this later?
+        if (identifier.startsWith("%permission:") || identifier.startsWith("<permission:")) { // Bad code, maybe fix this later?
             return permissionRefreshInterval;
         }
-        return refreshIntervals.getOrDefault(identifier, defaultInterval);
+        Integer interval = refreshIntervals.get(identifier);
+        if (interval != null) return interval;
+        String alternate = PlaceholderIdentifier.toPercentSyntax(identifier).equals(identifier)
+                ? PlaceholderIdentifier.toAngleBracketSyntax(identifier)
+                : PlaceholderIdentifier.toPercentSyntax(identifier);
```
NEZNAMY: Same question here.

shared/src/main/java/me/neznamy/tab/shared/placeholders/types/TabPlaceholder.java
Crafting-tunbol: At the current stage, it's needed because I didn't update the PlaceholderReplacementsConfiguration logic. But I agree it would be cleaner to force the correct syntax here. It would make more sense.

NEZNAMY: I don't think you'll need to change much. Also forcing %% for MP only in replacements sounds really confusing.

shared/src/main/java/me/neznamy/tab/shared/placeholders/PlaceholderRefreshConfiguration.java
Crafting-tunbol: Same answer here.

shared/src/main/java/me/neznamy/tab/shared/placeholders/PlaceholderIdentifier.java
```diff
@@ -0,0 +1,77 @@
...
+     * @param   identifier
+     *          identifier to convert
+     * @return  identifier in angle bracket syntax
+     */
+    @NotNull
+    public static String toAngleBracketSyntax(@NotNull String identifier) {
+        if (identifier.startsWith("%") && identifier.endsWith("%")) {
+            return "<" + identifier.substring(1, identifier.length() - 1) + ">";
+        }
+        return identifier;
```
NEZNAMY: These two functions now seem unused.

NEZNAMY:
Looking good now. Anything else before we unleash it to the world?

Crafting-tunbol:
All good for me!

NEZNAMY:
Thank you for your contribution!

PR #1686, opened 2026-07-07
Add opaque nametag visibility mode

Nattapat2871:
# Add opaque nametag visibility mode

## Summary

This pull request adds an additional nametag visibility mode for TAB:

- `/tab nametag opaque`
- `/tab nametag opaqueview`

The new mode keeps using Minecraft's existing scoreboard team nametag system. It does not introduce `text_display`, armor stands, or any custom floating nametag entities.

The goal is to allow nametags to stay visible normally, while preventing players from seeing nametags through blocks.

## Why

TAB currently supports showing and hiding nametags, but there is no built-in middle state for this behavior:

- visible when the viewer has direct line of sight
- hidden when a block obstructs the view

This PR adds that middle state as `opaque`.

### `/tab nametag opaque [player] [viewer] [-s]`

Enables opaque nametag visibility for a target player.

```mcfunction
/tab nametag opaque Steve
/tab nametag opaque Steve Alex
/tab nametag opaque Steve -s
```

### `/tab nametag opaqueview [viewer] [-s]`

Enables opaque nametag view mode for a viewer.

When enabled, the viewer sees all player nametags only when line of sight is clear. Players who join later are also handled by this viewer-level mode.

```mcfunction
/tab nametag opaqueview Alex
/tab nametag opaqueview Alex -s
```

### Target Mode

`/tab nametag opaque <player>` applies opaque visibility to the target player's nametag.

Other players can see that nametag only when there is no block obstruction.

### Viewer Mode

`/tab nametag opaqueview <viewer>` applies opaque visibility to everything the viewer sees.

The viewer sees all player nametags only when line of sight is clear.

## Existing Command Interaction

These commands clear opaque state where appropriate:

```text
/tab nametag show
/tab nametag hide
/tab nametag toggle
/tab nametag showview
/tab nametag hideview
/tab nametag toggleview
```

This avoids stale opaque state remaining active after switching back to normal show/hide behavior.

## Tab Completion

Tab completion was updated to include:

```text
opaque
opaqueview
```

Full command list:

```text
/tab nametag show
/tab nametag opaque
/tab nametag hide
/tab nametag toggle
/tab nametag showview
/tab nametag opaqueview
/tab nametag hideview
/tab nametag toggleview
```

## Platform Support

Line-of-sight checks were added for:

- Bukkit / Paper / Purpur
- Fabric
- Forge
- NeoForge

Proxy platforms keep the default behavior because block line-of-sight cannot be calculated there.

Folia keeps a safe fallback behavior because cross-entity line-of-sight checks are not safe from the global scheduler.

## Implementation Notes

This PR adds a new nametag invisibility reason:

```java
OPAQUE_OCCLUSION
```

Opaque nametag state is tracked separately from existing hide/show command state.

Line-of-sight checking is split into two phases:

```text
1. TAB's nametag thread snapshots player/viewer pairs.
2. The platform thread performs line-of-sight checks.
3. Results are applied back on TAB's nametag thread.
```

This avoids mutating TAB nametag state directly from the platform/server thread.

## Performance Notes

Opaque occlusion refresh runs periodically only when opaque mode is active.

If no players/viewers are using opaque mode, the refresh exits early.

The implementation avoids changing scoreboard teams unless the visibility state actually changes.

## What This PR Does Not Do

This PR does not:

- add custom rendered nametags
- add `text_display` nametags
- add armor stand nametags
- change prefix/suffix formatting
- change tablist formatting
- change chat formatting
- add persistent config storage for the command state
- add a new public API method

It only extends the existing `/tab nametag` command behavior.

## POV :  /tab nametag show 
https://github.com/user-attachments/assets/d74df3a1-e85d-4d3a-a436-6698fe3a9e79

## POV: If that player uses `/tab nametag opaque`, or you yourself use `/tab nametag opaqueview`
https://github.com/user-attachments/assets/849eb7eb-b97a-4697-8fd6-27179177373f
### `/tab nametag opaque`

Suggested demo:

```text
1. Run /tab nametag opaque <player>.
2. Look at the player's nametag with clear line of sight.
3. Move behind a block.
4. Show that the nametag is no longer visible through the block.
```

### `/tab nametag opaqueview`

Suggested demo:

```text
1. Run /tab nametag opaqueview <viewer>.
2. Show the viewer seeing nametags normally with clear line of sight.
3. Move behind a block.
4. Show that the viewer cannot see nametags through the block.
5. Let another player join after the command.
6. Show that the newly joined player's nametag also follows opaque view behavior.
```

## Verification

Tested successfully with:

```bash
./gradlew :shared:compileJava :bukkit:compileJava :velocity:compileJava :fabric:compileJava :forge:compileJava :neoforge:compileJava --stacktrace
./gradlew :shared:test --stacktrace
./gradlew :bukkit:shadowJar :bungeecord:shadowJar :velocity:shadowJar :fabric:jar :forge:jar :neoforge:jar --stacktrace
```

Also checked:

```bash
git diff --check
```

## Build Note

The aggregate build command:

```bash
./gradlew --no-configure-on-demand :jar:build --stacktrace
```

currently fails in the existing Paperweight setup for:

```text
:bukkit:paper_1_21_4:paperweightUserdevSetup
:bukkit:paper_1_21_9:paperweightUserdevSetup
```

The failure is caused by Paperweight/Codebook being executed with the repository's Java 25 toolchain and failing with:

```text
Unsupported class file major version 69
```

The repository already sets Java toolchain 25 in:

```text
build-logic/src/main/kotlin/tab.base-conventions.gradle.kts
```

This appears to be a build tooling / Paperweight compatibility issue and is not caused by the opaque nametag source changes in this pull request.

The source modules and primary platform artifacts listed above build successfully.

## Pull Request

```text
https://github.com/Nattapat2871/TAB/pull/new/feature/nametag-opaque-view
```

### DIFF
--- a/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/platform/BukkitPlatform.java
+++ b/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/platform/BukkitPlatform.java
@@ -389,6 +389,11 @@ public class BukkitPlatform implements BackendPlatform {
         Bukkit.getScheduler().runTask(plugin, task);
     }
 
+    @Override
+    public boolean hasLineOfSight(@NotNull TabPlayer viewer, @NotNull TabPlayer target) {
+        return ((Player) viewer.getPlayer()).hasLineOfSight((Player) target.getPlayer());
+    }
+
     /**
      * Converts component to string using bukkit RGB format if supported by the server.
      * If not, closest legacy color is used instead.
--- a/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/platform/FoliaPlatform.java
+++ b/bukkit/src/main/java/me/neznamy/tab/platforms/bukkit/platform/FoliaPlatform.java
@@ -171,4 +171,9 @@ public class FoliaPlatform extends BukkitPlatform {
         if (!getPlugin().isEnabled()) return; // Server shutdown, no one cares anymore, everyone is about to be kicked
         globalScheduler_execute.invoke(globalScheduler, getPlugin(), task);
     }
+
+    @Override
+    public boolean hasLineOfSight(@NotNull TabPlayer viewer, @NotNull TabPlayer target) {
+        return true; // Cross-entity line-of-sight checks are not safe from Folia's global scheduler.
+    }
 }
--- a/fabric/src/main/java/me/neznamy/tab/platforms/fabric/FabricPlatform.java
+++ b/fabric/src/main/java/me/neznamy/tab/platforms/fabric/FabricPlatform.java
@@ -274,6 +274,16 @@ public record FabricPlatform(MinecraftServer server) implements BackendPlatform 
     }
 
     @Override
+    public void runSyncGlobal(@NotNull Runnable task) {
+        server.execute(task);
+    }
+
+    @Override
+    public boolean hasLineOfSight(@NotNull TabPlayer viewer, @NotNull TabPlayer target) {
+        return ((FabricTabPlayer) viewer).getPlayer().hasLineOfSight(((FabricTabPlayer) target).getPlayer());
+    }
+
+    @Override
     @NotNull
     public Object dump() {
         Map<String, Object> map = new LinkedHashMap<>();
--- a/forge/src/main/java/me/neznamy/tab/platforms/forge/ForgePlatform.java
+++ b/forge/src/main/java/me/neznamy/tab/platforms/forge/ForgePlatform.java
@@ -263,6 +263,16 @@ public record ForgePlatform(MinecraftServer server) implements BackendPlatform {
     }
 
     @Override
+    public void runSyncGlobal(@NotNull Runnable task) {
+        server.execute(task);
+    }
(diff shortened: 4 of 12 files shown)


NEZNAMY:
The code is looking good. It's also good that you made it command based to not affect users who don't want this. 

I was just wondering, is this something people actually want? I don't think I've ever received such request. When players sneak, their nametag is already invisible through walls. Does this have such a great added value?

Nattapat2871:
Thank you for merging it.

About the use case: Yes, this is a niche feature, but I think it has value for servers that want nametags to behave more like normal in-world visibility.

Sneaking only hides the nametag when the player chooses to sneak. This feature is different because it lets the server enforce "not visible through blocks" behavior even when the player is standing, walking, fighting, or moving normally.

The main use case is for survival / SMP / PvP-style servers where admins still want normal Minecraft scoreboard nametags, but do not want players to use nametags as a way to locate others behind walls.

Since it is command-based, it stays opt-in and should not affect existing users unless they explicitly use it. :)

PR #1693, opened 2026-07-18
[Fand] Update API dependency to 0.8.3

MC20018:
## What changed
- Update the Fand compile-only API dependency from `0.7.8+build.1` to `0.8.3+build.1`.
- Keep the platform implementation and packaged runtime behavior unchanged.

## Verification
- Confirmed the `0.8.3+build.1` Maven metadata, POM, and JAR are publicly available.
- Ran `./gradlew :fand:compileJava` successfully with Java 25.

### DIFF
--- a/fand/build.gradle.kts
+++ b/fand/build.gradle.kts
@@ -1,6 +1,6 @@
 dependencies {
     implementation(projects.shared)
-    compileOnly("io.fand:fand-api:0.7.8+build.1")
+    compileOnly("io.fand:fand-api:0.8.3+build.1")
 }
 
 tasks.compileJava {


NEZNAMY:
Not related to this PR, but I have a question. You have removed dependency shading and relocation for the module. How does that work exactly? I assume the platform jar is small without relocation, but what about the combined jar? Does it still have all dependencies and relocation? If so, what's the point of even doing it? It's not like anyone would compile the plugin and use the platform jar instead of the combined jar that's being distributed.

MC20018:
That is a fair question.

The Fand module is not intended to be installed as a standalone platform jar. Its shadowJar is an intermediate artifact consumed by the universal `:jar:shadowJar` task. The universal jar unpacks and merges the already-built platform shadow jars; it does not resolve or relocate their dependencies again.

The Fand module still shades and relocates the dependencies it embeds, including bStats, json-simple, Kyori Event, YamlAssist, and SnakeYAML.

The dependencies intentionally excluded from the Fand shadow jar are the larger or optional libraries such as MySQL, Delivery4j, RabbitMQ, Redis, and their transitive dependencies. Fand loads those through its plugin library mechanism using `fand/src/main/resources/libraries.json`. They are therefore provided beside the plugin by Fand's plugin classloader instead of being embedded in the universal TAB jar.

The Fand API itself is also `compileOnly`, because `io.fand.api` and the server-provided Adventure classes are supplied by Fand at runtime.

So the combined jar still contains the dependencies and relocations produced by `:fand:shadowJar`. It deliberately does not contain the dependencies delegated to Fand's library loader. The standalone platform jar is mainly useful as a build intermediate and for platform-specific compilation/testing, while the actual distributed artifact remains the universal TAB jar.

NEZNAMY:
But what's the point of excluding Redis and MySQL in the platform jar if the combined jar still has them?

MC20018:
You are right. I was describing the Fand intermediate shadow JAR in isolation and overlooked how the universal JAR is assembled.

The universal JAR merges the shadow JARs from all platforms, and the other platform JARs already include MySQL, Redis, Delivery4j, and their relocated classes. Therefore, excluding those dependencies only from :fand does not reduce the distributed universal JAR. With the current libraries.json setup, Fand may also download another copy unnecessarily.

So the Fand-specific exclusions do not provide a meaningful benefit for the artifact users actually install. I will clean this up in a follow-up by aligning Fand with the other platform modules and removing the redundant runtime library loading.

Thanks for pointing this out.

Issue #1684, opened 2026-07-01
Memory leak: Thousands of BukkitTabPlayer and me.neznamy.tab instances retained after player disconnect


ghost:
### Server version

1.21.11

### TAB version

6.1.0

### Heap dump

I can't send the file as it's over 10 GB.

### Uncleared object(s)

Memory leak detected: The heap dump shows nearly 7,000 instances of BukkitTabPlayer and thousands of other objects related to me.neznamy.tab remaining in the memory, despite there being only 150 players currently online.
<img width="1414" height="360" alt="Image" src="https://github.com/user-attachments/assets/1f4777f9-d014-43bb-a328-28e09ed3dae7" />
<img width="577" height="323" alt="Image" src="https://github.com/user-attachments/assets/072df3d8-e18d-426b-9638-c7d8a8bf8c62" />

### Additional info

As shown in screenshots, the volume of these orphaned instances is causing significant and unnecessary memory consumption. Please investigate why these objects are failing to clear upon player logout. I can provide the full heap dump file if you need further investigation.

### Checklist

- [x] I am running latest version of the plugin
- [ ] I have included download link of my heap dump
- [ ] I ticked all of the boxes without actually reading them
- [x] I have explained what I believe is wrong with the heap dump

NEZNAMY:
Please send the full heap dump. The plugin is too big to search for memory leaks manually.

ghost:
Would it be a problem if I sent this to you via Discord? If it's not an issue, could you provide your username?

NEZNAMY:
`neznamy`

NEZNAMY:
Turned out to be caused by `ultraspoofpro` plugin. It creates fake players with fake channels, but those channels never get cleared, so TAB's netty handler remains alive with its reference to the respective player (who is no longer "online").

PR #1698, opened 2026-07-22
[Nametags] Use weak keys for registered team maps to fix disconnected player leak

SenseiJu:
### Problem

`NameTagPlayerData.registeredTeams` and `registeredProxyTeams` are `HashMap`s keyed by the team owner (`TabPlayer` / `ProxyPlayer`). The keys are strong references, so a disconnected owner that is not removed from the map stays alive forever. Two paths leave stale entries:

- `NameTag#unregisterTeam(target)` only iterates **online** viewers, so viewers that are themselves offline never drop the target, and
- an owner can be re-registered by a refresh that fires **after** quit (the placeholder refresh dispatch guards on `isLoaded()`, which never flips on disconnect, rather than `isOnline()`).

Every viewer keeps one of these maps keyed by every target, so a single missed removal compounds to roughly O(players²) retained objects.

### Impact (production heap dump, Paper 26.1.2, ~130 players online)

- **12,467** `BukkitTabPlayer` on the heap; **12,338** had `online == false` — i.e. `FeatureManager#onQuit` → `markOffline()` had already run, so these are fully-disconnected players that were never released (only ~130 were actually connected).
- Those instances retained **~7.9 GB (~57%)** of a 14 GB heap, which had pushed G1 Old Gen to ~100% and TPS to ~6.5.
- Path-to-GC-roots on sampled ghosts showed them pinned as **keys in ~1,450 `HashMap`s owned by `NameTagPlayerData`**, with no other plugin in the reference chain. The dump held ~20M `HashMap` / ~47M `HashMap$Node` overall, consistent with the per-viewer×target team maps.

### Fix

Switch `registeredTeams` and `registeredProxyTeams` to `WeakHashMap`, matching the two sibling per-player maps already in this class (`opaqueNameTagViewers`, `nameTagInvisibilityReasonsRelational`).

- While an owner is online it stays strongly reachable via `TAB#data`, so its entry is retained exactly as before — no behavior change for connected players.
- Once an owner disconnects and TAB drops its other references, the weak key lets it be collected and the stale entry is removed automatically, so a missed or late unregister can no longer leak.

This is safe because the values are `String`s that never reference the key (so entries are actually collectable), the maps are only used via `put` / `remove` / `containsKey` (no iteration or `size()` reliance), and `NameTag` is `CustomThreaded` so access is single-threaded — the same assumptions the existing `WeakHashMap` fields in this class already rely on.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardImpl.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardImpl.java
@@ -134,8 +134,6 @@ public class ScoreboardImpl extends RefreshableFeature implements me.neznamy.tab
      *          Player to send this scoreboard to
      */
     public void addPlayer(@NonNull TabPlayer p) {
-        if (!p.isOnline()) return; // player is not online, cannot send scoreboard
-
         if (p.scoreboardData.activeScoreboard == this) return; // already registered
         p.scoreboardData.titleProperty = new Property(this, p, title);
         p.getScoreboard().registerObjective(
--- a/shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardImpl.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardImpl.java
@@ -134,6 +134,8 @@ public class ScoreboardImpl extends RefreshableFeature implements me.neznamy.tab
      *          Player to send this scoreboard to
      */
     public void addPlayer(@NonNull TabPlayer p) {
+        if (!p.isOnline()) return; // player is not online, cannot send scoreboard
+
         if (p.scoreboardData.activeScoreboard == this) return; // already registered
         p.scoreboardData.titleProperty = new Property(this, p, title);
         p.getScoreboard().registerObjective(
--- a/shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/nametags/NameTag.java
@@ -143,6 +143,7 @@ public class NameTag extends TabFeature implements NameTagManager, JoinListener,
     public void onQuit(@NotNull TabPlayer disconnectedPlayer) {
         onlinePlayers.removePlayer(disconnectedPlayer);
         unregisterTeam(disconnectedPlayer);
+        disconnectedPlayer.teamData.clearRegisteredTeams();
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


NEZNAMY:
Using WeakHashMap may result in the players being cleared on quit before TAB's onQuit runs, which will make team unregister code never run and players will still have the team of disconnected player registered. I was thinking about this and couldn't find any specific example what could break. Maybe just on reconnect the team will not be registered again with possibly new values because it's already registered, but this is unlikely to be experienced. I still don't like this solution.

Not too long ago we got #1684, which turned out to be caused by some fake player plugin that holds references to channels, which hold tab players due to tab's netty injection, which then hold other players because the map is not cleared for player who disconnected. Are you sure there is no other cause? If you look at the code, if a player leaves, their team is unregistered for all online players. The markOffline() has no impact here, as every feature has its own list of online players that's updated on join/quit call of that feature. So even after player is marked as offline, onQuit is called and code runs. 

This should get proper investigation as something is clearly not working as intended. This memory leak should not be happening with the current design.
What I am willing to do is modify onQuit() handler to clear the maps for player who left, deleting references to current online players instead of expecting GC to delete the player object.

`🤖 Generated with Claude Code` is a really scary part of this PR as well.

SenseiJu:
Generated with claude code was so I dont have to do the write up xD

I don't think moving to a WeakHashMap will affect other functionaility, especially your concern of "may result in the players being cleared on quit before TAB's onQuit runs" since the onQuits closure references the object at this point so its still strongly reachable and not eligible for collection.

If you think modifying onQuit is a better fix then I would be happy to test it as well to make sure its working as intended but so far I have deployed the current WeakHashMap change and it's worked flawlessly, I have not had any issues with TAB's other functionalities or any errors being printed to console so it seems fairly stable. We have had over 100k joins since I deployed the change.

NEZNAMY:
You are right about the reachability. I would still prefer not using WeakHashMap for this though. They tend to forget things they shouldn't (I had a user with such issue). You can make the onQuit change yourself (you will need to expose the maps) to clear the maps for the player who left and let me know if it works fine.

SenseiJu:
Tested this recently based on your suggestion, seems to work just as well. 

Also found another racey type leak in the scoreboard, due to the `delay-on-join-milliseconds` config option, if a player was to attempt a join and either quickly disconnect/crash/or kicked any other way, the task will still run and add the player to the scoreboard leaving a ghost TabPlayer once again. Simple check to see if they are online when adding sorted it.

Going to be running this over the next day to ensure its fully stable and I will circle back but if you have any more suggestions let me know and I will try address them.

shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardImpl.java
```diff
@@ -134,6 +134,8 @@ public ScoreboardImpl(@NonNull ScoreboardManagerImpl manager, @NonNull String na
      *          Player to send this scoreboard to
      */
     public void addPlayer(@NonNull TabPlayer p) {
+        if (!p.isOnline()) return; // player is not online, cannot send scoreboard
```
NEZNAMY: Can you move this check to the delayed task instead?

SenseiJu: Moved

NEZNAMY:
Thanks for the contribution.

PR #1712, opened 2026-08-12
Fix VSA downstream sidebar yielding race

### LINKED ISSUE #1711: [VelocityScoreboardAPI] Downstream sidebar is cancelled before TAB yields its proxy objective
## TAB version

6.1.3-SNAPSHOT, built from current `master` (`9304615`).

## Software causing compatibility issue

- Velocity 4.1.0-SNAPSHOT
- VelocityScoreboardAPI 2.x
- TAB-Bridge on the Paper backend
- Paper 1.21.11 backend

## Additional info

When a backend plugin assigns a Bukkit sidebar objective (for example, `hudungeon`), VelocityScoreboardAPI fires `ObjectiveEvent.Display`. TAB receives that event, but its Velocity hook queues `FeatureManager#onDisplayObjective` through `CPUManager#runTask`.

VelocityScoreboardAPI continues handling the same downstream display packet before that queued task runs. It still finds TAB's proxy `TAB-Scoreboard` in the sidebar slot and cancels the downstream packet because proxy objectives have priority. TAB later logs that the external objective was received and disables its own slotting, but the client keeps seeing TAB's sidebar because the downstream display packet has already been discarded.

Observed TAB debug line:

```text
Player HuanLan_2021 received scoreboard called hudungeon, disabling TAB's scoreboard slotting.
```

The expected result is that the backend sidebar becomes visible immediately and TAB's sidebar returns after the backend objective is removed.

Suggested fix: handle the VSA display event synchronously. For an external sidebar objective, mark it as `otherPluginScoreboard` and unregister TAB's proxy objective before the VSA handler checks whether the proxy sidebar slot is occupied. When that external objective is unregistered, synchronously rebuild TAB's highest applicable scoreboard.

I prepared a minimal patch against current master which changes only TAB's VelocityScoreboardAPI hook and `ScoreboardManagerImpl`; it preserves the existing asynchronous/generic packet path for all non-VSA platforms.

juinwei7:
Fixes #1711

## Cause

VelocityScoreboardAPI gives proxy objectives priority over downstream objectives. TAB queued VSA `ObjectiveEvent.Display` handling through `CPUManager`, so the VSA packet handler checked the proxy sidebar slot before TAB had yielded it and cancelled the downstream display packet.

## Changes

- Handle VSA display/unregister events synchronously in the Velocity hook.
- When a downstream sidebar is displayed, unregister TAB's proxy objective before VSA performs its proxy-slot priority check.
- When that downstream objective is removed, rebuild the highest applicable TAB scoreboard.
- Preserve the existing generic asynchronous packet path for all non-VSA platforms.

## Verification

- `JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-22.jdk/Contents/Home ./gradlew :velocity:shadowJar --offline --no-daemon`
- Build successful; the Velocity implementation is compiled for Java 21 (classfile major version 65).

The live server trace that motivated this change is included in #1711. Runtime validation on the affected proxy network is pending deployment of this build.

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardManagerImpl.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/scoreboard/ScoreboardManagerImpl.java
@@ -189,6 +189,34 @@ public class ScoreboardManagerImpl extends RefreshableFeature implements Scorebo
         }
     }
 
+    /**
+     * Handles a sidebar display sent through VelocityScoreboardAPI.
+     *
+     * <p>VelocityScoreboardAPI gives proxy objectives priority over downstream
+     * objectives. Its display event is therefore the last opportunity to remove
+     * TAB's proxy objective before the downstream display packet is cancelled.
+     * This method is intentionally called synchronously by the Velocity hook.</p>
+     *
+     * @param receiver
+     *          player receiving the packet
+     * @param slot
+     *          display slot
+     * @param objective
+     *          objective assigned to the slot, or {@code null}
+     */
+    public void onVelocityScoreboardDisplay(@NotNull TabPlayer receiver, int slot, @Nullable String objective) {
+        if (slot != Scoreboard.DisplaySlot.SIDEBAR.ordinal()) {
+            return;
+        }
+        if (objective == null) {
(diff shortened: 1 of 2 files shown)


NEZNAMY:
Thanks for the contribution.

PR #1714, opened 2026-08-17
[API] Add BelowName and PlayerListObjective managers

crackma:
Adds BelowNameManager and PlayerListObjectiveManager, exposing per-player control over the belowname and playerlist objectives: forced enable/disable, title, value, and (playerlist only) render type, plus reset().

### DIFF
--- a/shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
+++ b/shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
@@ -268,10 +268,6 @@ public class BelowName extends RefreshableFeature implements JoinListener, QuitL
     public void setScore(@NotNull TabPlayer viewer, @NotNull TabPlayer scoreHolder, int value, @NotNull String fancyValue) {
         if (viewer.belowNameData.disabled.get()) return;
         if (viewer.server != scoreHolder.server || viewer.world != scoreHolder.world) return; // Viewer definitely cannot see this player in game
-        if (Boolean.FALSE.equals(scoreHolder.belowNameData.forcedEnabled)) {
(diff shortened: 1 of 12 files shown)


NEZNAMY:
Good job on the initial concept, but I have a few requests to align it with other APIs:
* Make `setValue` accept `String` for placeholder support
* Add `setFancyValue(TabPlayer, String)` methods
* Add getters for `getCustom*(TabPlayer)` returning the custom value set via API
* Allow arguments to be nullable to reset them instead of having a reset method (this seems to already be done, but the reset method is now redundant and might be confusing)

shared/src/main/java/me/neznamy/tab/shared/features/belowname/BelowName.java
```diff
@@ -235,6 +268,10 @@ private void register(@NotNull TabPlayer player) {
     public void setScore(@NotNull TabPlayer viewer, @NotNull TabPlayer scoreHolder, int value, @NotNull String fancyValue) {
         if (viewer.belowNameData.disabled.get()) return;
         if (viewer.server != scoreHolder.server || viewer.world != scoreHolder.world) return; // Viewer definitely cannot see this player in game
+        if (Boolean.FALSE.equals(scoreHolder.belowNameData.forcedEnabled)) {
+            viewer.getScoreboard().removeScore(OBJECTIVE_NAME, scoreHolder.getNickname());
```
NEZNAMY: Keep in mind disabling is for viewer, not target players. This kind of thing isn't even possible before 26.2, and even after that isn't something TAB supports through this option. It is supported if the fancy-value evaluates to an empty string, but not like this.

NEZNAMY:
Thanks for the contribution.

PR #1722, opened 2026-08-29
[Plugin messages] Fix error when receiving an unknown message type from Bridge

### LINKED ISSUE #1721: PluginMessageDecodeTask indexes registeredMessages with an unchecked readByte()
**File:** `shared/src/main/java/me/neznamy/tab/shared/task/PluginMessageDecodeTask.java`

```java
Supplier<IncomingMessage> supplier = registeredMessages[in.readByte()];
```

`registeredMessages` holds 11 entries. `readByte()` is signed, so any byte that is negative or
`>= 11` throws `ArrayIndexOutOfBoundsException`; an empty payload makes `readByte()` itself throw.
In practice this happens when the Bridge and the plugin are on mismatched versions.

The surrounding `TimedCaughtTask` catches it and reports it through the error manager, so nothing
crashes — but the message is dropped and the log gets an exception where a plain diagnostic would
be clearer.

**Suggested fix**, five lines, matching the shape TAB already uses on its other channel
(`ProxySupport` + `ErrorManager#unknownProxyMessage`):

```java
int type = in.readByte();
if (type < 0 || type >= registeredMessages.length) {
    TAB.getInstance().getErrorManager().printError(
        "Received a plugin message of unknown type (" + type + "): incompatible Bridge version?", null);
    return;
}
Supplier<IncomingMessage> supplier = registeredMessages[type];
```

LordEris:
Fixes #1721.

`PluginMessageDecodeTask` uses a byte read off the wire directly as an array index:

```java
Supplier<IncomingMessage> supplier = registeredMessages[in.readByte()];
```

`registeredMessages` holds 11 entries, so any type outside `0-10` throws
`ArrayIndexOutOfBoundsException`. `TimedCaughtTask` catches it, so nothing breaks — but `errors.log`
gets a raw stack trace that says nothing about the cause, and it repeats for every message received.

This is defensive validation of a network-sourced index, not a version-mismatch fix. `tab:bridge-6`
already pins the major version, so a Bridge from another major line never reaches this code. What the
channel does not cover is a minor mismatch inside 6.x — a Bridge build that added a type — or any
other backend plugin writing on `tab:bridge-6`, which the proxy has no control over.

The type is now bounds-checked and reported through `ErrorManager`, mirroring what `ProxySupport`
already does for an unknown proxy action (`ProxySupport.java:82-84` +
`ErrorManager#unknownProxyMessage`). It is read unsigned, so a sender writing `200` is logged as
`200` rather than `-56`.

The message is deliberately factual — it states the type received and the range this build knows,
without guessing at a cause it cannot determine.

Still present on `master` (da8befc) and on tag 6.1.2. Found by reading the code; no user report known.

**Scope:** this covers the type byte only. An empty or truncated payload still fails through
`taskThrewError`, since `msg.read(in)` is unguarded — that is a separate concern, and the consistent
fix there would be a `try`/`catch` around decoding as in `ProxySupport.processMessage`. Happy to
follow up with that if you want it.

**Not built or run locally:** I had no environment able to resolve the Gradle dependencies. The change
is two files, mirrors an existing pattern, and adds no new imports.


NEZNAMY:
> What the channel does not cover is a minor mismatch inside 6.x — a Bridge build that added a type

>  An empty or truncated payload still fails through taskThrewError, since msg.read(in) is unguarded — that is a separate concern, and the consistent fix there would be a try/catch around decoding as in ProxySupport.processMessage

See https://github.com/NEZNAMY/TAB/issues/1721#issuecomment-5460906941

> or any
other backend plugin writing on tab:bridge-6, which the proxy has no control over.

It is not TAB's responsibility to try to prevent malicious plugins from being malicious. There is a million things they can do and they are much "better" than doing this.

> mirroring what ProxySupport already does for an unknown proxy action

That's because ProxySupport dynamically registers messages based on enabled features, so this can actually happen if enabled features mismatch. Not an ideal solution, but that's how it works and that's why it's different from bridge communication.

PR #1723, opened 2026-08-29
[MySQL] Make /tab mysql upload atomic so a failed insert cannot empty a table

### LINKED ISSUE #1720: /tab mysql upload can leave tab_groups/tab_users empty when an insert fails (no transaction)
**File:** `shared/src/main/java/me/neznamy/tab/shared/command/MySQLCommand.java`

`upload()` deletes both tables and then re-inserts:

```
DELETE FROM tab_groups
DELETE FROM tab_users
  ... then re-insert every group and user
```

The connection runs with JDBC autocommit (`MySQL#openConnection` never calls
`setAutoCommit(false)`, and no `commit()` / `rollback()` appears anywhere in the module). Each
`DELETE` is therefore committed on its own, before any row is written back.

If an insert fails — a lost connection, a value the server rejects — the deletes have already
landed. `tab_groups` or `tab_users` is then left partially populated, or empty, instead of keeping
the configuration that was there before. The operator gets an error and a table that no longer
holds their data.

**Suggested fix**, roughly twelve lines across two files: wrap the whole upload in one
transaction. Add `beginTransaction()`, `commit()` and a quiet `rollback()` next to
`MySQL#execute`, and have `MySQLCommand#upload` begin before the first delete, commit after the
last insert, and roll back in a `catch`.

**Two traps worth flagging to whoever implements it**, because a naive `setAutoCommit(false)`
would give a false guarantee:

1. `prepareStatement` reconnects silently when the connection has dropped. A reconnect in the
   middle of a transaction silently starts a *new* one — the rollback then protects nothing.
   The transaction helpers need to fail loudly instead of reconnecting mid-transaction.
2. The in-memory maps are updated alongside the SQL. A rollback restores the tables but not the
   maps, leaving memory and database disagreeing until the next reload.

LordEris:
Fixes #1720.

`MySQLCommand#upload` clears both tables, then re-inserts groups first and users second:

```java
mysql.execute("DELETE FROM tab_groups");
mysql.execute("DELETE FROM tab_users");
upload(groupFile, ...getGroups());
upload(userFile,  ...getUsers());
```

The connection runs with JDBC auto-commit — `MySQL#openConnection` never calls `setAutoCommit(false)`,
and no `commit()` / `rollback()` exists in the module — so each `DELETE` is committed on its own,
before any row is written back. If an insert throws, the deletes have already landed: the `catch`
reports "MySQL upload failed" and leaves the table empty instead of keeping what was there.

### Steps to reproduce

`tab_users` is declared `value varchar(1024)` and MySQL runs in strict mode by default, so an
oversized value is a deterministic insert failure — no privilege changes, no timing race.

1. Set `mysql.enabled: true` with a working connection.
2. Run `/tab mysql upload` once so both tables hold data. Confirm both `SELECT COUNT(*)` are non-zero.
3. In `users.yml`, give any user a `tabprefix` of **1100 characters**.
4. Run `/tab mysql upload` again.

**Result:** the command reports "MySQL upload failed". `tab_groups` is repopulated — its inserts ran
first and succeeded — but `SELECT COUNT(*) FROM tab_users` returns **0**. The previous per-user
configuration is gone, and nothing warned it would be.

**Expected:** either the whole upload applies, or none of it does and both tables keep what they had.

### The fix

The upload runs in one transaction, rolled back in the existing `catch`.

`prepareStatement` reopened a lost connection on its own. Left as is, a reconnect mid-transaction
would silently start a *new* one — the statements already sent stay committed and the rollback
protects nothing — so it now throws while a transaction is open instead of reconnecting.

**Known limit, not addressed here:** the in-memory maps are written alongside the SQL, so after a
rollback memory holds the new values while the database holds the old, until the next reload.
Reloading from the database after a rollback would close that; I left it out to keep this change to
the atomicity problem. Happy to add it if you prefer it in the same PR.

**Not built or run locally:** I had no environment able to resolve the Gradle dependencies.


NEZNAMY:
Did you actually run into this problem? You solution does not truly resolve the issue, only tries to reduce the damage. Since the uploading process uploads data from files, it means the content in the files is safe, contains everything the database has / should have, and it is ready to be re-uploaded after resolving this issue. Why would anyone configure prefix longer than 1000 characters anyway? If it's for gradients, using any supported gradient syntax over using manual colors drastically reduces the length. Even if not and every letter has a different color, that's like 10 characters per visible character, which means your visible prefix would still be 100 characters long, which no one would want anyway.

I will most likely be closing this pull request as it doesn't even solve the linked issue. 

Together with your other pull request I have a suspicion you are desperately looking for problems that don't exist and just want to become a contributor. I apologize for the accusation if not.

LordEris:
I'm not looking to become a contributor; it's just something I noticed while reviewing a fork I created using a bot. No worries—I'm closing this one too.

